PR #61, opened 2016-06-13
correctly dispose rootDisposable

yole:
Calling Disposable.dispose() does not dispose other objects that depend on it. Disposer.dispose() must be used to ensure the entire Disposable tree is disposed.


cbeams:
Hi @yole, could you sign our CLA (Individual or Company as you see fit) at http://gradle.org/contributor-license-agreement/#? We'll merge this afterward, thanks.

Issue #72, opened 2016-06-15
Identify `buildscript` block failure cases and otherwise ensure parity with Groovy


cbeams:
Gradle's `buildscript` block requires special case handling, both in Groovy- and Kotlin-based scripts. The `buildscript` block must be discovered and handled in a "first pass" of the build script in order to calculate the classpath that should be used during the "second pass" in which the rest of the build script is processed.

This epic is about identifying as many failure cases as possible around `buildscript` blocks, and ensuring that we handle them correctly, provide proper error messages, and so forth. For example:
- What happens if more than one `buildscript` block is defined within in a single script?
- What happens if more than one `buildscript` block is defined across scripts that are processed together using `applyFrom`?
- What happens if a `buildscript` block is declared within a conditional?

See #68 for a concrete example of one of these issues.
See #73 for an example of how to submit one of these issues as a pull request with a failing JUnit test.

Please add similar issues or pull requests for any failure use cases you can find.

PR #73, opened 2016-06-15
Guard against multiple buildscript blocks in a single script

cbeams:
See failing test case. Part of #72.

### DIFF
--- a/src/test/kotlin/org/gradle/script/lang/kotlin/integration/GradleScriptKotlinIntegrationTest.kt
+++ b/src/test/kotlin/org/gradle/script/lang/kotlin/integration/GradleScriptKotlinIntegrationTest.kt
@@ -7,6 +7,7 @@ import org.gradle.testkit.runner.GradleRunner
 
 import org.junit.Rule
 import org.junit.Test
+import org.junit.rules.ExpectedException
 import org.junit.rules.TemporaryFolder
 
 import java.io.File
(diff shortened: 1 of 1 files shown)


bamboo:
This doesn't seem to match the behaviour of the Groovy frontend which is to consider all `buildscript` blocks in the first step.

cbeams:
I just did a very quick test on the Groovy side, and I thought it was discarding all but the first `buildscript { }`. If that's not the case, then we should adjust this, including the test case.

bamboo:
Check this one out, it will only compile if you uncomment the `dependencies` block and at least one of the `repositories` blocks:
 
```
buildscript {
    //repositories { mavenCentral() }
}

buildscript {
    //repositories { jcenter() }
}

buildscript {
    //dependencies { classpath("junit:junit:4.12") }
}

org.junit.Assert.assertTrue(true)

task dumpRepositories {
    doLast {
        buildscript.repositories.each {
            println(it.url)
        }
    }
}
```

oehme:
I think only allowing one `buildscript` block at the very top of the build file is a good idea. It's a very slight departure from Groovy, but if we were to redesign it, that's what we would do. We have done the same for the `plugins` block already.

bamboo:
Resolved in 588ab800a0aae6c379e258938ef28af605295fb7.

PR #299, opened 2017-03-10
Generate dependency extension for kotlinx modules.

sureshg:
Added a handy dependency extension for kotlinx modules (specifically for coroutines).


JLLeitschuh:
@bamboo This PR makes sense to add. Merge conflicts need to be resolved though. Thoughts?

bamboo:
Thanks for the PR, however we think this particular method would be better in the `kotlin-gradle-plugin` or similar.

Issue #302, opened 2017-03-14
Setting non-existent Kotlin build script in settings.gradle should not fail


pioterj:
Setting non-existent build script file with `.kts` extension results in `settings.gradle` like:

`rootProject.buildFileName = 'build.non-existent.kts'`

results in build failing with:

`kotlin.KotlinNullPointerException (no error message)`

It should not fail as build scripts are optional.

eskatos:
If the file doesn't exists, Gradle provides a `NonExistentFileScriptSource`.
The Groovy provider compiles the text provided by this `ScriptSource`, which is `""`.
The Kotlin provider works with `scriptSource.resource.file` which is `null` in this case, causing the NPE.

bamboo:
I think there are two different use cases here:

 1. For a project, the default build script (automatically detected or defined in `settings.gradle`) is optional and so we shouldn't throw if the file isn't there
 2. For a script plugin, e.g. `apply from: "someScript.gradle.kts"`, we should throw if the file isn't there

That reasoning tells me we shouldn't be fixing _1_ at the script plugin factory level but at a higher-level.

eskatos:
The Gradle runtime already distinguishes the various use cases and fails with a proper error message when it has to.

A scripting provider may decide to add stuff to build scripts, e.g. base plugins like we envision in order to solve #320. Even when the file isn't there because it is optional.

So, I think this boils down to the contracts of `NonExistentFileScriptSource` and it's companion `EmptyFileTextResource` being loosely defined.

eskatos:
When it comes to the settings scripts there actually are three cases:
1. specified settings script file exists
2. specified settings script file does not exists
3. using empty settings, there's no `File` involved, see `SettingsLocation` and `BuildLayout` in GBT

In the 1st case the `ScriptSource` passed to the `ScriptPluginFactory` is a `UriScriptSource`, already properly handled.

In the 2nd case it is a `NonExistentFileScriptSource` whose `TextResource` returns a `null` `File`. We need to accommodate here.

In the 3rd case the `ScriptSource` passed to the `ScriptPluginFactory` is a `StringScriptSource` that use its class name as `fileName`. In that case we can't distinguish the "file" path and name.

So, the `KotlinScriptPluginFactory` must not use the nullable `file` property of the `ScriptSource#resource` in order to handle all those cases while letting the Gradle runtime make the decisions wrt. optional scripts and eventual failures.

bamboo:
[...]

PR #331, opened 2017-04-13
Do not fail on non-existing or “empty” script compilation

eskatos:
Do not consume ScriptSource TextResource nullable file property.
Rework the multi-kotlin-project-config-injection sample to demonstrate
a subproject with no build script.

See #302

### DIFF
--- a/samples/multi-kotlin-project-config-injection/build.gradle.kts
+++ b/samples/multi-kotlin-project-config-injection/build.gradle.kts
@@ -16,8 +16,16 @@ allprojects {
     }
 }
 
-// Configure all KotlinCompile tasks on each sub-project
+// Apply and configure the Kotlin Gradle plugin on each sub-project
 subprojects {
+
+    apply {
+        plugin("kotlin")
+    }
+
+    dependencies {
+        compile(kotlinModule("stdlib"))
+    }
 
     tasks.withType<KotlinCompile> {
         println("Configuring $name in project ${project.name}...")
--- a/src/main/kotlin/org/gradle/script/lang/kotlin/provider/CachingKotlinCompiler.kt
+++ b/src/main/kotlin/org/gradle/script/lang/kotlin/provider/CachingKotlinCompiler.kt
@@ -64,7 +64,7 @@ class CachingKotlinCompiler(
         return compileScript(cacheKeyPrefix + scriptFileName + buildscript, classPath, parentClassLoader) { cacheDir ->
             ScriptCompilationSpec(
                 KotlinBuildscriptBlock::class,
-                scriptFile,
+                scriptFile.path,
                 cacheFileFor(buildscript, cacheDir, scriptFileName),
                 scriptFileName + " buildscript block")
         }
@@ -83,7 +83,7 @@ class CachingKotlinCompiler(
         val compiledScript = compileScript(cacheKeyPrefix + scriptFileName + plugins, classPath, parentClassLoader) { cacheDir ->
             ScriptCompilationSpec(
                 KotlinPluginsBlock::class,
-                scriptFile,
+                scriptFile.path,
                 cacheFileFor(plugins, cacheDir, scriptFileName),
                 scriptFileName + " plugins block")
         }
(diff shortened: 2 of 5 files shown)


samples/multi-kotlin-project-config-injection/build.gradle.kts
```diff
@@ -16,9 +16,17 @@ allprojects {
     }
 }
 
-// Configure all KotlinCompile tasks on each sub-project
+// Apply and configure the Kotlin Gradle plugin on each sub-project
 subprojects {
 
+    apply {
```
bamboo: We should make sure `cli/build.gradle.kts` can still be edited with the correct classpath when this sample is imported in IntelliJ.

src/main/kotlin/org/gradle/script/lang/kotlin/provider/KotlinBuildScriptCompiler.kt
```diff
@@ -59,9 +59,9 @@ class KotlinBuildScriptCompiler(
     val gradleApiExtensions: ClassPath,
     gradleScriptKotlinJars: ClassPath) {
 
-    val scriptResource = scriptSource.resource!!
-    val scriptFile = scriptResource.file!!
-    val script = convertLineSeparators(scriptResource.text!!)
+    val scriptFilePath = scriptSource.fileName!!
+    val scriptFile = File(scriptFilePath)
```
bamboo: From what I understood, `scriptSource.fileName` doesn't necessarily point to an existing file, right?

If that's the case we shouldn't be creating a `File` out of it as it might be used in cache keys later on (see `compilePluginsBlockOf`).

src/main/kotlin/org/gradle/script/lang/kotlin/provider/CachingKotlinCompiler.kt
```diff
@@ -94,16 +94,17 @@ class CachingKotlinCompiler(
...
     fun compileBuildScript(
         scriptFile: File,
+        script: String,
         additionalSourceFiles: List<File>,
         classPath: ClassPath,
         parentClassLoader: ClassLoader): CompiledScript {
 
         val scriptFileName = scriptFile.name
-        return compileScript(cacheKeyPrefix + scriptFileName + scriptFile, classPath, parentClassLoader) {
+        return compileScript(cacheKeyPrefix + scriptFileName + script, classPath, parentClassLoader) { cacheDir ->
```
bamboo: This behaves slightly different as `File` and `String` have [different cache key semantics](https://github.com/gradle/gradle/blob/93a68297faf3681679e58253bd73e326f02fa9e4/subprojects/core/src/main/java/org/gradle/cache/internal/DefaultCacheKeyBuilder.java#L63-L63).

I imagine it might have a (very small) performance impact when the same build script file is reused many times in a single build.

Just an observation.

src/main/kotlin/org/gradle/script/lang/kotlin/provider/KotlinBuildScriptCompiler.kt
eskatos: The `File` instance here is just a vehicule for "path" and "filename". It isn't used as a *file*.
When the file doesn't exists, it carries its intended path.

Given that we require Java 8 we could use `Path` instead to better convey the intention. But

src/main/kotlin/org/gradle/script/lang/kotlin/provider/CachingKotlinCompiler.kt
eskatos: Yep, performance wise the difference is that before we hashed the file from disk and now we hash its content, that was already loaded in any case, from memory.

samples/hello-world/settings.gradle
```diff
@@ -1 +1 @@
-rootProject.buildFileName = 'build.gradle.kts'
+rootProject.buildFileName = 'buildZog.gradle.kts'
```
eskatos: Oops, fixed 😄

eskatos:
After discussing this with @bamboo, we decided step back and fix that once we migrated to the new Kotlin embedded compiler API.

Issue #320, opened 2017-03-31
Remove `ProjectExtensionsTaskRegistrationAction` configuration overhead


bamboo:
Execution cost should only be paid by Kotlin based projects.

An alternative implementation would be to introduce a service that reacts to Kotlin build script evaluation events and registers the tasks as needed.

bamboo:
A better strategy would be to apply a plugin to the root project from `KotlinScriptPlugin`. That plugin would register the `gskGenerateAccessors` task.

eskatos:
This has been merged

PR #332, opened 2017-04-13
Remove ProjectExtensionsTaskRegistrationAction, apply a plugin instead

eskatos:
That way, no need to check for .kts files presence.
Builds not using gsk won’t suffer the performance penalty anymore.
Builds using gsk will benefit a small performance gain.

See #320


eskatos:
With this approach, a multi-project build with its root project using a Groovy build script and subprojects using Kotlin build scripts, the root project won't get the `gskGenerateAccessors` task.

src/main/kotlin/org/gradle/script/lang/kotlin/provider/KotlinScriptBasePlugin.kt
```diff
@@ -0,0 +1,44 @@
...
+    override fun apply(project: Project) {
+        if (project.rootProject == project) {
+            createGenerateAccessorsTask(project)
+        }
+        createDisplayAccessorsTask(project)
+    }
+
+    private
+    fun createGenerateAccessorsTask(project: Project)
+        = with(project) {
```
bamboo: Please keep the `=` in the same line as the function declaration and let's standardise on `project.run` instead of `with(project)`

src/main/kotlin/org/gradle/script/lang/kotlin/provider/KotlinScriptPlugin.kt
```diff
@@ -39,6 +39,10 @@ class KotlinScriptPlugin(
         if (target !is Project) {
             throw IllegalArgumentException("target $target was not a Project as expected")
         }
+        with(target) {
+            if (!plugins.hasPlugin(KotlinScriptBasePlugin::class.java))
+                plugins.apply(KotlinScriptBasePlugin::class.java)
```
bamboo: `PluginManager#apply(Class<*>)` is idempotent so `hasPlugin` is unnecessary here.

eskatos:
> This PR doesn't implement the exact same behaviour as before. In particular, a multi-project build with a root Groovy project and a Kotlin sub-project should still expose `gskGenerateAccessors`.

Yep, let's discuss this today.

eskatos:
@bamboo could you have another look?

PR #346, opened 2017-04-28
Integration test samples

eskatos:
This PR introduce a `samplesTest` source set with integration tests for all samples.
The task rule to run samples has been removed, favoring the integration tests.

Code shared by tests moved to a `fixtures` source set.


eskatos:
Once this is merged we need to change our CI job for samples to run `samplesTest` instead of `checkSamples`.

PR #347, opened 2017-04-29
Reuse implicit imports provided by Gradle and add Kotlin specifics

eskatos:
Confirmed that it works by removing imports in samples.
Once this is in, we'll be able to remove imports for Gradle public types from our samples and the GSK build script.

### DIFF
--- a/src/main/kotlin/org/gradle/script/lang/kotlin/resolver/KotlinBuildScriptModelBuilder.kt
+++ b/src/main/kotlin/org/gradle/script/lang/kotlin/resolver/KotlinBuildScriptModelBuilder.kt
@@ -25,6 +25,7 @@ import org.gradle.internal.classpath.DefaultClassPath
 
 import org.gradle.script.lang.kotlin.accessors.accessorsClassPathFor
 import org.gradle.script.lang.kotlin.provider.KotlinScriptClassPathProvider
+import org.gradle.script.lang.kotlin.support.ImplicitImports
 import org.gradle.script.lang.kotlin.support.serviceOf
 
 import org.gradle.tooling.provider.model.ToolingModelBuilder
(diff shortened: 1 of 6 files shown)


src/main/kotlin/org/gradle/script/lang/kotlin/resolver/KotlinBuildScriptModelBuilder.kt
```diff
@@ -35,6 +36,7 @@ import java.io.Serializable
 
 interface KotlinBuildScriptModel {
     val classPath: List<File>
+    val implicitImports: List<String>
```
bamboo: I'd move it after `sourcePath`.

src/main/kotlin/org/gradle/script/lang/kotlin/support/ImplicitImports.kt
```diff
@@ -16,11 +16,10 @@
...
-        "org.gradle.api.*",
-        "org.gradle.api.plugins.*",
-        "org.gradle.script.lang.kotlin.*",
+import org.gradle.configuration.ImportsReader
 
-        "java.io.File")
-}
+class ImplicitImports(
+    importsReader: ImportsReader,
+    val list: List<String> = importsReader.importPackages.map { "$it.*" }.plus(listOf(
```
bamboo: I'd keep the list of imports as a lazy field and introduce two explaining functions.

```
class ImplicitImports(private val importsReader: ImportsReader) {

    val list by lazy {
        gradleImports() + gradleScriptKotlinImports()
    }

    fun gradleImports() = importsReader.importPackages.map { "$it.*" }

    fun gradleScriptKotlinImports() = listOf("org.gradle.script.lang.kotlin.*", "java.io.File")
}
```

PR #348, opened 2017-05-02
Tighten exported API

eskatos:
This commit adds `internal` or `private` to internal members
The only exported API that includes Gradle internals is now in .provider

### DIFF
--- a/src/main/kotlin/org/gradle/script/lang/kotlin/cache/BuildServices.kt
+++ b/src/main/kotlin/org/gradle/script/lang/kotlin/cache/BuildServices.kt
@@ -22,10 +22,10 @@ import org.gradle.cache.CacheRepository
 import org.gradle.cache.internal.CacheKeyBuilder
 
 
+internal
 object BuildServices {
 
     @Suppress("unused")
-    private
     fun createScriptCache(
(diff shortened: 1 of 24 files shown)


src/main/kotlin/org/gradle/script/lang/kotlin/cache/BuildServices.kt
```diff
@@ -22,10 +22,10 @@ import org.gradle.cache.CacheRepository
 import org.gradle.cache.internal.CacheKeyBuilder
 
 
+internal
 object BuildServices {
 
     @Suppress("unused")
-    private
```
bamboo: Interesting. I had the impression factory methods could not be public.

src/main/kotlin/org/gradle/script/lang/kotlin/concurrent/tapi.kt
```diff
@@ -30,13 +30,15 @@ import kotlin.coroutines.experimental.suspendCoroutine
  *
  * Execution will continue on the TAPI executor thread.
  */
+internal
```
bamboo: We've been using the convention of writing`internal inline` on the same line.

PR #358, opened 2017-05-09
TAPI model builder for script template classpath

eskatos:
The model type should end up in a separate jar but with this we can already start trying this out.


eskatos:
Superseded by e773d0d15942a72bb7fe28192b157480ceaa37cf

Issue #386, opened 2017-06-07
`embedded-kotlin` plugin


eskatos:
- applies the kotlin plugin from the portal using the `embeddedKotlinVersion`
- adds pinned dependency to `kotlin-stdlib` and `kotlin-reflect` at `embeddedKotlinVersion`

Solution must include strategy for publishing new releases via TeamCity.

eskatos:
Publishing from CI was extracted as #396

PR #392, opened 2017-06-09
Introduce embedded-kotlin plugin

eskatos:
See #386

This PR does not include publishing from TC, this will be addressed sideways.


bamboo:
Merged manually in a7fe3474b1930ec68ed4deaa33aac9cedf9e5867.

PR #412, opened 2017-07-07
Gradle Plugin Sample taking advantage of `kotlin-dsl`

eskatos:


samples/gradle-plugin/build.gradle.kts
```diff
@@ -0,0 +1,14 @@
+tasks {
+
+    val publishPlugin by creating(GradleBuild::class) {
+        dir = file("plugin")
+        tasks = listOf("publish")
+    }
+
+    val checkSample by creating(GradleBuild::class) {
+        dir = file("sample")
```
bamboo: ❌ I think `consumer` would be a better name. What do you think?

samples/gradle-plugin/build.gradle.kts
```diff
@@ -0,0 +1,14 @@
+tasks {
+
+    val publishPlugin by creating(GradleBuild::class) {
+        dir = file("plugin")
+        tasks = listOf("publish")
+    }
+
+    val checkSample by creating(GradleBuild::class) {
```
bamboo: :x: The task in the sample should make sense on its own. I think `consumePlugin` would be a better name, see below.

eskatos:
Thanks @bamboo for the feedback.
Comments addressed, could you have another look?

samples/gradle-plugin/build.gradle.kts
```diff
@@ -0,0 +1,14 @@
+tasks {
+
+    val publishPlugin by creating(GradleBuild::class) {
+        dir = file("plugin")
+        tasks = listOf("publish")
+    }
+
+    val checkConsumer by creating(GradleBuild::class) {
```
bamboo: I'm a bit bothered by the `check` prefix here.

How about renaming `publishPlugin` to just `plugin` and `checkConsumer` to just `consumer`?

eskatos: Sounds good.
Done

PR #477, opened 2017-09-01
Add Spotless plugin for code linting and reformatting

JLLeitschuh:
### Context

Any project that has many contributors should have a consistent style guide and style guide linter.
Lacking such tends to lead to diffs with white space changes all over because developers have different editors with different settings.

This adds kotlin code linting with [Ktlint](https://ktlint.github.io/) currently the only code linter/formatter for kotlin that I know about.

Ktlint is driven with the [Spotless](https://github.com/diffplug/spotless/tree/master/plugin-gradle) plugin that simply checks to see if running the various formatters you configure on it will create a diff. If a diff will happen then the spotless task fails.
By default this plugin runs source linting as part of the `check` phase. However, for build file linting you need to explicitly call the `spotlessCheck` task.

Now, there is no more need to setup your editor in the same way as other developers.
Simply run `./gradlew spotlessApply` before committing and spotless will make your code Spotless!

### Contributor Checklist

### Notes
Although this looks like a large diff it's really not. Mostly imports and some minor indentation things.
If you are interested in this PR try to get it in before other changes get pulled in as there is a higher likelihood for merge conflicts.

### DIFF
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/accessors/AccessorsClassPath.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/accessors/AccessorsClassPath.kt
@@ -18,31 +18,23 @@ package org.gradle.kotlin.dsl.accessors
 
 import org.gradle.api.Project
 import org.gradle.api.internal.project.ProjectInternal
-
 import org.gradle.cache.internal.CacheKeyBuilder.CacheKeySpec
-
 import org.gradle.internal.classpath.ClassPath
 import org.gradle.internal.classpath.DefaultClassPath
-
 import org.gradle.kotlin.dsl.cache.ScriptCache
 import org.gradle.kotlin.dsl.codegen.fileHeader
 import org.gradle.kotlin.dsl.support.compileToJar
 import org.gradle.kotlin.dsl.support.loggerFor
 import org.gradle.kotlin.dsl.support.serviceOf
-
 import org.jetbrains.kotlin.utils.addToStdlib.firstNotNullResult
-
 import org.jetbrains.org.objectweb.asm.ClassReader
 import org.jetbrains.org.objectweb.asm.Opcodes.ACC_PUBLIC
 import org.jetbrains.org.objectweb.asm.Opcodes.ACC_SYNTHETIC
-
 import java.io.BufferedWriter
 import java.io.Closeable
 import java.io.File
-
-import java.util.*
+import java.util.AbstractMap
 import java.util.jar.JarFile
-
 
 fun accessorsClassPathFor(project: Project, classPath: ClassPath) =
     project.getOrCreateSingletonProperty {
@@ -50,9 +42,7 @@ fun accessorsClassPathFor(project: Project, classPath: ClassPath) =
             ?: AccessorsClassPath(ClassPath.EMPTY, ClassPath.EMPTY)
     }
 
-
 data class AccessorsClassPath(val bin: ClassPath, val src: ClassPath)
-
 
 private
 fun buildAccessorsClassPathFor(project: Project, classPath: ClassPath) =
@@ -67,11 +57,9 @@ fun buildAccessorsClassPathFor(project: Project, classPath: ClassPath) =
             DefaultClassPath(accessorsSourceDir(cacheDir)))
     }
 
-
 private
 fun configuredProjectSchemaOf(project: Project) =
     aotProjectSchemaOf(project) ?: jitProjectSchemaOf(project)
-
 
 private
 fun aotProjectSchemaOf(project: Project) =
@@ -80,7 +68,6 @@ fun aotProjectSchemaOf(project: Project) =
         .getOrCreateSingletonProperty { multiProjectSchemaSnapshotOf(project) }
         .schema
         ?.let { it[project.path] }
-
 
 private
 fun jitProjectSchemaOf(project: Project) =
@@ -91,10 +78,8 @@ fun jitProjectSchemaOf(project: Project) =
         schemaFor(project).withKotlinTypeStrings()
     }
 
-
 private
 fun scriptCacheOf(project: Project) = project.serviceOf<ScriptCache>()
-
 
 private
 fun buildAccessorsJarFor(projectSchema: ProjectSchema<String>, classPath: ClassPath, outputDir: File) {
@@ -115,13 +100,11 @@ fun buildAccessorsJarFor(projectSchema: ProjectSchema<String>, classPath: ClassP
     })
 }
 
-
 internal
 sealed class TypeAccessibility {
     data class Accessible(val type: String) : TypeAccessibility()
     data class Inaccessible(val type: String, val reasons: List<InaccessibilityReason>) : TypeAccessibility()
 }
-
 
 internal
 sealed class InaccessibilityReason {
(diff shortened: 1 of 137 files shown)


provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptBasePlugin.kt
```diff
@@ -24,16 +24,14 @@ import org.gradle.kotlin.dsl.accessors.tasks.UpdateProjectSchema
 import org.gradle.kotlin.dsl.apply
 import org.gradle.kotlin.dsl.task
 
-
 class KotlinScriptBasePlugin : Plugin<Project> {
-    override fun apply(project: Project): Unit =
+    override fun apply(project: Project) : Unit = // ktlint-disable
```
JLLeitschuh: I discovered a small bug:
https://github.com/shyiko/ktlint/issues/77

provider/src/main/kotlin/org/gradle/kotlin/dsl/accessors/AccessorsClassPath.kt
```diff
@@ -115,40 +100,36 @@ fun buildAccessorsJarFor(projectSchema: ProjectSchema<String>, classPath: ClassP
...
 
-
 internal
 sealed class InaccessibilityReason {
 
     data class NonPublic(val type: String) : InaccessibilityReason()
     data class NonAvailable(val type: String) : InaccessibilityReason()
     data class Synthetic(val type: String) : InaccessibilityReason()
 
-    val explanation get() = when (this) {
```
JakeWharton: I would also call this a bug. When only a `get` function is supplied for a property this shorthand should be preferred.

JLLeitschuh: This may just be that there isn't a check for this. I ran the intellij formatter to get rid of most of the problems with star imports that Ktlint can't fix for you. Both code patterns are valid according to the formatter.

The way the tool was designed was such that it doesn't make decisions about format for things that haven't been decided upon by the community here: https://github.com/yole/kotlin-style-guide

Let me revert this format change. Sorry about that.

bamboo:
This is promising, @JLLeitschuh, thanks.

Can it be configured to handle the other conventions we're using such as `two empty lines between module level members` and `align when branches`?

JLLeitschuh:
> Can it be configured to handle the other conventions we're using such as `two empty lines between module level members` and `align when branches`?

Sorry, no. Ktlint is sold as:

> An anti-bikeshedding Kotlin linter with built-in formatter

The only configuration options that it offers is the indentation level and character (tabs vs. spaces). You can see the [README here](https://github.com/shyiko/ktlint) where the author describes the decision.

These options are only available by adding a `.editorconfig` file to the root directory of the project.

In case you are unfamiliar with the term [bikeshedding](http://bikeshed.com/) (reload the page a few times once you're done (you'll understand the joke)).

JLLeitschuh:
I almost forgot about spotless on windows. Hence the addition of the `. gitattributes` file. This PR should be good to go as far as I know.

JLLeitschuh:
I can also configure spotless to add a license header to all source files too if that's something you're interested in.

That could also be a follow-on PR.

JLLeitschuh:
Let me amend my previous statement regarding adding your own rules.

Yes, it can be done by adding your own custom `RuleSets` but then you have to write something that adds those checks and that really starts to enter the realm of bikeshedding. In my mind the best way is to just go with the rules that exist and open bugs when you discover one.

The author of Ktlint is really responsive and if its a fast fix might have a release for you several hours after you report a problem.

JLLeitschuh:
Okay, I'm happy with this PR as is. I don't think I need to add anything else.

Unless you want me to add configuration to spotless to add licence headers to files (I'd rather do that in a follow on PR).

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptBasePlugin.kt
eskatos: ❌ This PR should be updated to use 0.9.2 that fixes this bug before being merged.

provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/SourcePathProvider.kt
```diff
@@ -54,7 +53,6 @@ object SourcePathProvider {
     private
     fun subDirsOf(dir: File): Collection<File> =
         if (dir.isDirectory)
-            dir.listFiles().filter { it.isDirectory }
-        else
+            dir.listFiles().filter { it.isDirectory } else
```
eskatos: ❌  This change looks off to me

bamboo:
> I went through the diff and I was surprised by how little changes there are.

Yeah, given that we've been very consistent with style and that the suggested changes in style are not particularly appealing, I will need more time to consider this PR. Let's revisit this after the `0.12.0` cycle.

Thank you very much for you effort so far, Jonathan!

provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/SourcePathProvider.kt
JLLeitschuh: This is from the IntelliJ settings that ktlint suggests you use in your IDE but is not required.
It may be in a future release of ktlint.

JLLeitschuh:
Would you like me to hold off on updating this PR until after `0.12.0`? Are there other things you'd like to see? Would it be better to drop this PR?
Personally, I don't care about styling, I just want it to be consistent so that I don't have to ensure that I have my editor configured the same as everyone else.

--- 

RE: habits that choke with existing style

I can't really do anything about these differences. I don't think that there are currently any other formatters that we could embed in the build for kotlin at the moment.

It would be nice if IntelliJ shipped their formatter as an independent artifact that spotless could depend upon.

JLLeitschuh:
I found an alternative that may satisfy your desires.
https://github.com/arturbosch/detekt
I don't really have the to go through and set it up though.

bamboo:
Thanks for the link, @JLLeitschuh. `detekt` looks promising.

eskatos:
Took a look at `detekt`.
Lots of configuration possibilities 🚲 
One blocker as of my quick investigation https://github.com/arturbosch/detekt/issues/457

eskatos:
https://github.com/arturbosch/detekt/issues/457 is now fixed and will be part of the next detekt release

After this, the only thing we do that cannot be configured is the alignment of `when` clauses.
Writing a custom detekt rule looks feasible but ... 🚲 
It cannot be configured in IntelliJ either and must be done manually anyway.

So, I propose that we drop the habit of aligned `when` clauses in favor of easier style compliance for contributions.

WDYT @bamboo?

bamboo:
> It cannot be configured in IntelliJ either and must be done manually anyway.

IntelliJ does it automatically if you set `Code Style / Kotlin / Wrapping and Braces / Align 'when' branches in columns`.

eskatos:
Ah cool, I missed this one.
So, iff we wanna introduce `detekt` we'd need a rule for this.
It would also make sense to share IntelliJ setup in the repository.

bamboo:
Superseded by #677.

Issue #470, opened 2017-08-23
Feature: new sample hello-kotlin-js


guenhter:
Like the `hello-android` project it would be very helpful to have a `hallo-java-script` because with the actual gradle 4.1 release I'm not able to get this right any more.

This should be covered:
https://kotlinlang.org/docs/tutorials/javascript/getting-started-gradle/getting-started-with-gradle.html

eskatos:
Hi @guenhter, thanks for the issue.
What are the issues you're having?
We'd like to have such a sample, would you consider contributing it?

guenhter:
Hi, I'd love to contribute it, but due to my lack of understanding of new kotlin-dsl, I failed to do so. In my repo https://github.com/guenhter/kotlin-noob I already have a couple of projects (3 JVM based and one JS based) but the JS-based one is not longer working (I guess since gradle 4.1).

eskatos:
@guenhter, I opened a small pull-request against your repository that fix the build script for your `kotlin-js` module, see the comments there: https://github.com/guenhter/kotlin-noob/pull/1

guenhter:
@eskatos Thanks for the PR. I updated some compiler options in that project and I guess this could be used as JavaScript sample (if you want to use it).

eskatos:
Sure.

Here is some guidance that I hope will be helpful:
- create a `samples/hello-kotlin-js-world` directory
- populate it with a Gradle Kotlin DSL project, maybe start from the existing `samples/hello-kotlin-world`
- tweak the new sample build to produce Javascript

We'll then need to find a way to actually test this sample in order to prevent any regression.
I need to think about this, we don't want to fire a browser for the sole purpose of testing this.

You can use the following command to validate that your sample pass some basic tests:
```shell
./gradlew samples-test:test --tests="*SamplesSmokeTest"
```

Feel free to start on the pull-request!
Please remember to work from the `develop` branch of this repository.

Thanks!

eskatos:
This has been delivered by #478 
Closing

PR #478, opened 2017-09-04
Add JavaScript sample

guenhter:
### Context

New sample which targets another language: JavaScript

https://github.com/gradle/kotlin-dsl/issues/470

### Contributor Checklist


samples/hello-js/build.gradle.kts
```diff
@@ -0,0 +1,45 @@
...
+    "compile"(kotlin("stdlib-js"))
+}
+
+repositories {
+    jcenter()
+}
+
+tasks.withType<Kotlin2JsCompile> {
+    kotlinOptions.outputFile = "$projectDir/web/output.js"
+    kotlinOptions.moduleKind = "plain"
```
eskatos: :x: `plain` is the default value, this line should be removed

samples/hello-js/build.gradle.kts
```diff
@@ -0,0 +1,45 @@
...
+dependencies {
+    "compile"(kotlin("stdlib-js"))
+}
+
+repositories {
+    jcenter()
+}
+
+tasks.withType<Kotlin2JsCompile> {
+    kotlinOptions.outputFile = "$projectDir/web/output.js"
```
eskatos: :x: build output should be moved below `$buildDir` instead

samples/hello-js/build.gradle.kts
```diff
@@ -0,0 +1,45 @@
...
+    kotlinOptions.moduleKind = "plain"
+    kotlinOptions.sourceMap = true
+}
+
+val build by tasks
+build.doLast {
```
eskatos: :x: this should be modeled as a separate task and wired to `assemble` instead of `build`, e.g. :

```kotlin
val stageWeb by tasks.creating
stageWeb.doLast {
    // ...
}
val assemble by tasks
assemble.dependsOn(stageWeb)
```

samples/hello-js/build.gradle.kts
```diff
@@ -0,0 +1,45 @@
...
+    configurations["compile"].forEach { file: File ->
```
eskatos: :x: please use the `compileClasspath` configuration instead

samples/hello-js/README.md
```diff
@@ -0,0 +1,10 @@
+Hello, JavaScript!
+==================
+
+This sample demonstrates a simple JavaScript application with a Kotlin based build script.
+
+Simply import the Gradle project by pointing to the `setting.gradle` file.
+
+Build the project with `./gradlew clean build` to create the JavaScript-File (output is folder `web`).
+
+Open the `index.html` and check the Browser's console for output.
```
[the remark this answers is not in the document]
guenhter: The `index.html` is currently stored directly in the sample's root folder, so `./index.html`. Do you see this problematic?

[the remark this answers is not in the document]
guenhter: I'd move the `index.html` from the sample's root dir into `src/main/resources`. To copy it into `build/web/` I'd configure `the<JavaPluginConvention>().sourceSets["main"].output.resourcesDir = File("$buildDir/web/")`. It seems that I have to use `the<JavaPluginConvention>()` and not just `java`... What do you think about this?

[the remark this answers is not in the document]
eskatos: Moving it to `src/main/resources` and adding it to the `stageWeb` copy task should be enough.

No need to configure any source set.

BTW, you can't use the `java` static accessor here because the Kotlin JS plugin is not applied using the `plugins {}` block, see https://youtrack.jetbrains.com/issue/KT-20156

samples/hello-js/build.gradle.kts
```diff
@@ -0,0 +1,52 @@
...
+            includeEmptyDirs = false
+
+            from(zipTree(file.absoluteFile))
+            into("$buildDir/web")
+            include { fileTreeElement ->
+                val path = fileTreeElement.path
+                path.endsWith(".js") && (path.startsWith("META-INF/resources/") || !path.startsWith("META-INF/"))
+            }
+        }
+        copy {
```
eskatos: :x: This copy statement should be moved outside the `forEach` loop

samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/JavaScriptSampleTest.kt
```diff
@@ -0,0 +1,39 @@
...
+        // Wire `console.log` to `print`
+        engine.eval("var console = {}; console.log = print;")
+
+        // Load Kotlin JS stdlib
+        val pathToKotlinJsStdLib = "$projectRoot/build/web/kotlin.js"
+        val pathToBuildOutputJs = "$projectRoot/build/web/output.js"
+
+        engine.eval(FileReader(File(pathToKotlinJsStdLib)))
+        // Run build output
+        engine.eval(FileReader(File(pathToBuildOutputJs)))
```
eskatos: :o: The two comment lines above looks misplaced

guenhter:
@eskatos Thank's for your comments. This is very helpful. I added the requested changes.

eskatos:
@guenhter thanks for your work!
I amended your PR before merging, see 008c078d30f9f7a6bd0bb1a313570cf83d4caf6c and 18a0b44e98dd6187fb41b4c6230f11173c318fc3.

PR #482, opened 2017-09-05
Add maven-publish example and tests.

shradej1:
### Context

Gradle project has a maven-publish sample.  I was unable to find any existing examples of how to configure publications using Kotlin DSL.

### Contributor Checklist

I tried to add a source jar to the publication as well, but when I did, the default jar artifact (containing .class files) seemed to get replaced by a jar containing the sources.  I'm still trying to figure out if that was my mistake, or if it's a bug.


eskatos:
Adding a published source jar in the sample is a good idea 👍 

Here's the sample build script enhanced with it:

```kotlin
import org.gradle.api.tasks.bundling.Jar

plugins {
    `maven-publish`
    kotlin("jvm")
}

group = "org.gradle.sample"
version = "1.0.0"

repositories {
    jcenter()
}

dependencies {
    compile(kotlin("stdlib"))
}

val sourcesJar by tasks.creating(Jar::class) {
    classifier = "sources"
    from(java.sourceSets["main"].allSource)
}

publishing {
    repositories {
        maven {
            // change to point to your repo, e.g. http://my.org/repo
            url = uri("$buildDir/repo")
        }
    }

    publications.create("mavenJava", MavenPublication::class.java) {
        from(components["java"])
        artifact(sourcesJar)
    }
}
```

shradej1:
Great, thanks!  I'll update the pull request.

My first attempt at adding the source jar was almost identical to your
suggestion, although I was setting the 'classifier' in the artifact
configuration closure:

import org.gradle.api.tasks.bundling.Jar

plugins {
    `maven-publish`
    kotlin("jvm")
}

group = "org.gradle.sample"
version = "1.0.0"

repositories {
    jcenter()
}

dependencies {
    compile(kotlin("stdlib"))
}
val sourcesJar by tasks.creating(Jar::class) {
    from(java.sourceSets["main"].allSource)
}

publishing {
    repositories {
        maven {
            // change to point to your repo, e.g. http://my.org/repo
            url = uri("$buildDir/repo")
        }
    }

    publications.create("mavenJava", MavenPublication::class.java) {
        from(components["java"])
        artifact(sourcesJar, {
            classifier = "sources"
        })
    }
}


That seemed to more closely match the example in
https://docs.gradle.org/current/userguide/publishing_maven.html.  Although,
that approach appears to unintentionally replace the default artifact in
the publication with the sourcesJar.  The result is that the sources jar is
published once as maven-publish-1.0.0.jar, and again as
maven-publish-1.0.0-sources.jar.  Unless I'm missing something, that seems
like a bug.

I may have a chance this weekend to try and track down what's going on, if
that's useful.

shradej1:
Great, thanks! I'll update the pull request. My first attempt at adding the source jar was almost identical to your suggestion, although I was setting the 'classifier' in the artifact configuration closure: 

```
import org.gradle.api.tasks.bundling.Jar

plugins {
    `maven-publish`
    kotlin("jvm")
}

group = "org.gradle.sample"
version = "1.0.0"

repositories {
    jcenter()
}

dependencies {
    compile(kotlin("stdlib"))
}

val sourcesJar by tasks.creating(Jar::class) {
    from(java.sourceSets["main"].allSource)
}

publishing {
    repositories {
        maven {
            // change to point to your repo, e.g. http://my.org/repo
            url = uri("$buildDir/repo")
        }
    }

    publications.create("mavenJava", MavenPublication::class.java) {
        from(components["java"])
        artifact(sourcesJar, {
            classifier = "sources"
        })
    }
}
```

That seemed to more closely match the example in https://docs.gradle.org/current/userguide/publishing_maven.html. Although, that approach appears to unintentionally replace the default artifact in the publication with the sourcesJar. The result is that the sources jar is published once as maven-publish-1.0.0.jar, and again as maven-publish-1.0.0-sources.jar. Unless I'm missing something, that seems like a bug. I may have a chance this weekend to try and track down what's going on, if that's useful.

JLLeitschuh:
This pull request should be targeted at the `develop` branch. Not master. You may need to rebase your changes for that to work.

shradej1:
Rebased and targeted to develop

samples/maven-publish/README.md
```diff
@@ -0,0 +1,5 @@
+maven-publish
+==========
+
+Demonstrates how to apply and configure the `maven-publish` plugin to publish a kotlin library with accompanying
+sources.
```
eskatos: :x: Should add some guidelines on how to run the sample (`publish` task) and where the sample repository is.

samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/MavenPublishSampleTest.kt
```diff
@@ -0,0 +1,39 @@
...
+            Assert.assertTrue(numClassEntries > 0)
+        }
+
+        // repo contains source file
+        mavenPath.resolve("$projectName-$projectVersion-sources.jar").toFile().let { sourceFile ->
+            Assert.assertTrue("source file $sourceFile exists", sourceFile.exists())
+
+            // source file contains java and kt files
+            Assert.assertTrue(JarFile(sourceFile).stream().anyMatch { it.name.endsWith(".java") })
+            Assert.assertTrue(JarFile(sourceFile).stream().anyMatch { it.name.endsWith(".kt") })
```
eskatos: :x: Please use static imports for assertion methods

samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/MavenPublishSampleTest.kt
```diff
@@ -0,0 +1,39 @@
...
+        // then:
+        val repo = projectRoot.toPath().resolve("build").resolve("repo")
+        val mavenPath = repo.resolve(Paths.get("org", "gradle", "sample", projectName, projectVersion))
+
+        // repo contains jar file
+        mavenPath.resolve("$projectName-$projectVersion.jar").toFile().let { jarFile ->
+            Assert.assertTrue("jar file $jarFile exists", jarFile.exists())
+
+            // jar file contains .class entries
+            val numClassEntries = JarFile(jarFile).stream().filter { it.name.endsWith(".class") }.count()
```
eskatos: :x: `JarFile` instances should be closed, can use `JarFile(jarFile).use { .. }`

samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/MavenPublishSampleTest.kt
```diff
@@ -0,0 +1,39 @@
...
+import java.util.jar.JarFile
+
+class MavenPublishSampleTest : AbstractSampleTest("maven-publish") {
+    @Test
+    fun `publish`() {
+        val projectName = "maven-publish"
+        val projectVersion = "1.0.0"
+
+        // when:
+        build("publish")
```
bamboo: The test should avoid messing up with `~/.m2` by setting `maven.repo.local` as it's done in https://github.com/gradle/kotlin-dsl/blob/dcd84adf60591726b276fb644c6795dab86d85db/samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/MavenPluginSampleTest.kt#L13

shradej1: Done.  I'm trying to understand the concern though.  Are there instances when `publish` pushes artifacts to ~/.m2?  I thought that only happened with `publishToMavenLocal`.

JLLeitschuh: That is a pretty good question. I'd love to know the answer too.

eskatos: You're right @shradej1, `publish` does not trigger `publishToMavenLocal`.

Setting `maven.repo.local` is not necessary here. Did I miss something @bamboo?

bamboo: I assumed `publish` had similar behavior to `uploadArchives` with regards to the `maven-metadata-remote.xml` file but that doesn't seem to be the case. The property is indeed unnecessary.

shradej1: Thanks for the clarification.  Should I remove it or leave it as is?

samples/maven-publish/README.md
shradej1: Done.

samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/MavenPublishSampleTest.kt
eskatos: I just manually confirmed that `publish` does not mess with `~/.m2/`.
@shradej1 yes, please, remove it

shradej1: @eskatos reverted maven repo change and rebased.

samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/MavenPublishSampleTest.kt
```diff
@@ -0,0 +1,43 @@
...
+
+        // then:
+        val repo = projectRoot.toPath().resolve("build").resolve("repo")
```
eskatos: :x: Should use `existing("build/repo")` instead

samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/MavenPublishSampleTest.kt
```diff
@@ -0,0 +1,43 @@
...
+        val mavenPath = repo.resolve(Paths.get("org", "gradle", "sample", projectName, projectVersion))
```
eskatos: :x: Should use `existing("$repo/org/gradle/sample/$projectName/$projectVersion")` instead

eskatos:
Thank you @shradej1!

Issue #291, opened 2017-03-05
Unify `DependencyHandler` configuration dependency APIs for both extension methods and for the string invocation


mkobit:
## Overview

The generated extensions for various configurations have different method signatures than those when using the `String.invoke` and are also not easily visible when coding against them in IntelliJ. This leads to a different user experience when dealing with each type. such as not being able to do `"myConfiguration"("group", "name", "version")`

This issue is a proposal to unify the different dependency handling APIs.

## Per configuration generated extensions methods

For generated extensions methods for configurations like `compile`, there is:

* `compile(dependency: T, dependencyConfiguration: T.() -> kotlin.Unit): T`
* `compile(dependencyNotation: kotlin.Any): org.gradle.api.artifacts.Dependency`
* `compile(group: kotlin.String, name: kotlin.String, version: kotlin.String? = null, configuration: kotlin.String? = null, classifier: kotlin.String? = null, ext: kotlin.String? = null): org.gradle.api.artifacts.ExternalModuleDependency`
* `compile(group: kotlin.String, name: kotlin.String, version: kotlin.String? = null, configuration: kotlin.String? = null, classifier: kotlin.String? = null, ext: kotlin.String? = null, dependencyConfiguration: org.gradle.api.artifacts.ExternalModuleDependency.() -> kotlin.Unit): org.gradle.api.artifacts.ExternalModuleDependency`
* `compile(dependencyNotation: kotlin.String, dependencyConfiguration: org.gradle.api.artifacts.ExternalModuleDependency.() -> kotlin.Unit): org.gradle.api.artifacts.ExternalModuleDependency`

and from IntelliJ it looks like

<img width="1154" alt="compile extension methods on Dependency Handler" src="https://cloud.githubusercontent.com/assets/1447148/23584088/29c31bba-011c-11e7-94e1-1e263f01dc12.png">

## `String.invoke`

For `String.invoke` there is:

* `<T : org.gradle.api.artifacts.ModuleDependency> kotlin.String.invoke(dependency: T, dependencyConfiguration: T.() -> kotlin.Unit): T`
* `String.invoke(dependencyNotation: kotlin.Any): org.gradle.api.artifacts.Dependency`
* `String.invoke(dependencyNotation: kotlin.String, dependencyConfiguration: org.gradle.api.artifacts.ExternalModuleDependency.() -> kotlin.Unit): org.gradle.api.artifacts.ExternalModuleDependency`

[...]

PR #486, opened 2017-09-07
Unify all DependencyHandler Configuration extensions

eskatos:
So that

- generic static extensions,
- `String.invoke` extensions for dynamic Configuration access,
- and extensions generated according to the Project model

all provide the same API.

Also add javadoc to all extensions.

See #291

### DIFF
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/DependencyHandlerScope.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/DependencyHandlerScope.kt
@@ -39,9 +39,76 @@ class DependencyHandlerScope(val dependencies: DependencyHandler) : DependencyHa
     operator fun String.invoke(dependencyNotation: Any): Dependency =
         dependencies.add(this, dependencyNotation)
 
+    /**
+     * Adds a dependency to the given configuration.
+     *
+     * @param dependencyNotation notation for the dependency to be added.
+     * @param dependencyConfiguration expression to use to configure the dependency.
+     * @return The dependency.
+     * @see DependencyHandler.add
+     */
+    inline
     operator fun String.invoke(dependencyNotation: String, dependencyConfiguration: ExternalModuleDependency.() -> Unit): ExternalModuleDependency =
         dependencies.add(this, dependencyNotation, dependencyConfiguration)
 
+    /**
+     * Adds a dependency to the given configuration.
+     *
+     * @param group the group of the module to be added as a dependency.
+     * @param name the name of the module to be added as a dependency.
+     * @param version the optional version of the module to be added as a dependency.
+     * @param configuration the optional configuration of the module to be added as a dependency.
+     * @param classifier the optional classifier of the module artifact to be added as a dependency.
+     * @param ext the optional extension of the module artifact to be added as a dependency.
+     * @return The dependency.
+     *
+     * @see DependencyHandler.add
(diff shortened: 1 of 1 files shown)


provider/src/main/kotlin/org/gradle/kotlin/dsl/DependencyHandlerScope.kt
```diff
@@ -39,9 +39,76 @@ class DependencyHandlerScope(val dependencies: DependencyHandler) : DependencyHa
...
+     * @param configuration the optional configuration of the module to be added as a dependency.
+     * @param classifier the optional classifier of the module artifact to be added as a dependency.
+     * @param ext the optional extension of the module artifact to be added as a dependency.
+     * @param dependencyConfiguration expression to use to configure the dependency.
+     * @return The dependency.
+     *
+     * @see DependencyHandler.create
+     * @see DependencyHandler.add
+     */
+    inline
```
JLLeitschuh: Why is this method inline where the others are not?

provider/src/main/kotlin/org/gradle/kotlin/dsl/DependencyHandlerScope.kt
```diff
@@ -39,9 +39,76 @@ class DependencyHandlerScope(val dependencies: DependencyHandler) : DependencyHa
...
+    operator fun String.invoke(
+        group: String,
+        name: String,
+        version: String? = null,
+        configuration: String? = null,
+        classifier: String? = null,
+        ext: String? = null,
+        dependencyConfiguration: ExternalModuleDependency.() -> Unit): ExternalModuleDependency =
```
JLLeitschuh: Why do you need the overload above? Why not just make the default value for this param `{ }`?
Also, shouldn't this param be marked `noinline`? I'm surprised the compiler isn't blowing up on you for not having that keyword.

JLLeitschuh:
Lovely to see the `Configuration` overloads too!

provider/src/main/kotlin/org/gradle/kotlin/dsl/DependencyHandlerScope.kt
eskatos: This method is not the only one that is not inlined. It's not a higher-order function so it doesn't need to be, see https://kotlinlang.org/docs/reference/inline-functions.html

provider/src/main/kotlin/org/gradle/kotlin/dsl/DependencyHandlerScope.kt
eskatos: Good point, I unified the extensions according to what we generate for discovered configurations. I'll have a closer look, maybe we can reduce the number of extensions.

bamboo: This wasn't possible at the time this overload was introduced. Makes sense to add a default value to the argument (assuming all current usages of the method are supported).

eskatos: `e: ../org/gradle/kotlin/dsl/DependencyHandlerScope.kt: (99, 9): 'dependencyConfiguration: ExternalModuleDependency.() -> Unit = {}' construction is not yet supported in inline functions`

So, I'll keep the overloads for now.

Issue #483, opened 2017-09-06
Implicit imports comply with the Gradle core public API definition


eskatos:
Comply with Gradle core definition of the default imports by disambiguating simple type names amongst types with same simple name in different Gradle API packages.

For example, current wildcard implicit imports fail to resolve the `Jar` type as it matches both `org.gradle.api.tasks.bundling.Jar` and `org.gradle.jvm.tasks.Jar` ; forcing users to use explicit imports.

The Gradle internal `ImportsReader` service provides the metadata required to solve this. The Kotlin DSL provider should use `ImportsReader.simpleNameToFullClassNamesMapping` instead of
`ImportsReader.importPackages`. Moreover, he later will be removed from Gradle internals at some point.

See https://github.com/gradle/gradle/pull/2876 for some discussion around this.

eskatos:
Closing, see a1c39bb81410afc11b51c02b1b16b00159057272

PR #491, opened 2017-09-12
Implicit imports comply with the Gradle core public API definition

eskatos:
Comply with Gradle core definition of the default imports by
disambiguating simple type names amongst types with same simple name in
different Gradle API packages.

Use ImportsReader.simpleNameToFullClassNamesMapping instead of
ImportsReader.importPackages.

See #483

### DIFF
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/ImplicitImports.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/ImplicitImports.kt
@@ -29,7 +29,8 @@ class ImplicitImports(private val importsReader: ImportsReader) {
     }
 
     private
-    fun gradleImports() = importsReader.importPackages.map { "$it.*" }
+    fun gradleImports() =
+        importsReader.simpleNameToFullClassNamesMapping.values.map { it.first() }
 
     private
     fun gradleKotlinDslImports() =
--- a/provider/src/test/kotlin/org/gradle/kotlin/dsl/support/ImplicitImportsTest.kt
+++ b/provider/src/test/kotlin/org/gradle/kotlin/dsl/support/ImplicitImportsTest.kt
@@ -1,11 +1,9 @@
 package org.gradle.kotlin.dsl.support
 
 import org.gradle.kotlin.dsl.fixtures.AbstractIntegrationTest
-import org.hamcrest.CoreMatchers.equalTo
+import org.hamcrest.CoreMatchers.containsString
 import org.junit.Assert.assertThat
 import org.junit.Test
-import java.io.File
-import java.util.jar.JarFile
 
 class ImplicitImportsTest : AbstractIntegrationTest() {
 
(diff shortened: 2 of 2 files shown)


provider/src/test/kotlin/org/gradle/kotlin/dsl/support/ImplicitImportsTest.kt
```diff
@@ -0,0 +1,42 @@
...
+    @Test
+    fun `implicit imports are fully qualified to allow use of the preferred type amongst those with same simple name in different Gradle API packages`() {
+
+        // given:
+        existing("settings.gradle").appendText("""
+            rootProject.name = "some-project"
+        """)
+        withBuildScript("""
+
+            plugins {
```
bamboo: :x: This test could be both simpler and more precise by asserting on the resolved type name and only that with something like:

```kotlin
println("*${Jar::class.qualifiedName}*")
```

And then checking the output for`*org.gradle.api.tasks.bundling.Jar*`.

provider/src/test/kotlin/org/gradle/kotlin/dsl/support/ImplicitImportsTest.kt
```diff
@@ -0,0 +1,42 @@
...
+                java
+            }
+
+            // Prefer org.gradle.api.tasks.bundling.Jar over org.gradle.jvm.tasks.Jar
+            tasks.withType<Jar> {
```
JLLeitschuh: Do you want to add something like `assert(this is org.gradle.api.tasks.bundling.Jar)` just to be explicit?

Also, what is the difference between `org.gradle.api.tasks.bundling.Jar` and `org.gradle.jvm.tasks.Jar`?

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/ImplicitImports.kt
```diff
@@ -29,7 +29,8 @@ class ImplicitImports(private val importsReader: ImportsReader) {
     }
 
     private
-    fun gradleImports() = importsReader.importPackages.map { "$it.*" }
+    fun gradleImports() =
+        importsReader.simpleNameToFullClassNamesMapping.values.map { it.first() }
```
bamboo: 👍

provider/src/test/kotlin/org/gradle/kotlin/dsl/support/ImplicitImportsTest.kt
eskatos: Done!

provider/src/test/kotlin/org/gradle/kotlin/dsl/support/ImplicitImportsTest.kt
eskatos: > Also, what is the difference between org.gradle.api.tasks.bundling.Jar and org.gradle.jvm.tasks.Jar?

[`org.gradle.jvm.tasks.Jar`](https://github.com/gradle/gradle/blob/685d9df6b2437908c49ba6e2fd54ed8c6c7ff0ac/subprojects/platform-jvm/src/main/java/org/gradle/jvm/tasks/Jar.java) is the real task

[org.gradle.api.tasks.bundling.Jar](https://github.com/gradle/gradle/blob/a3c1bd72dc1fbee660b6b7da8cbf57e1758c83ac/subprojects/plugins/src/main/java/org/gradle/api/tasks/bundling/Jar.java) is a remnant of past refactoring, here for backwards compatibility

I mixed up the two in the comments that can be seen the line above your comment, sorry about the confusion.

PR #455, opened 2017-08-09
Add JDK 9 to travis build matrix

JLLeitschuh:
### Context

Related/demonstration of to #454

### Contributor Checklist
- [x] Base the PR against the `develop` branch
- [x] [Sign Gradle CLA](http://gradle.org/contributor-license-agreement/)
- [x] Provide integration tests to verify changes from a user perspective
- [x] Provide unit tests to verify logic
- [ ] ~Ensure that tests pass locally: `./gradlew check --parallel`~ (This is the problem I'm trying to demonstrate)

Issue #454, opened 2017-08-09
KT-20167: Kotlin DSL project won't build under JDK 9


JLLeitschuh:
The oracle JDK 9 is due to be released September 23. Many more developers will have JDK 9 on their systems. As such, the Kotlin DSL needs to be able to support being built and tested under JDK 9.

### Expected Behavior

I would expect the ability to build and test the Kotlin DSL build to succeed under JDK 9.

### Current Behavior

Currently, in my testing on Travis CI the build fails:
https://travis-ci.org/JLLeitschuh/kotlin-dsl/jobs/262691730#L487-L490

<details>
<summary>Very long log</summary>
<pre>
$ javac -J-Xmx32m -version
javac 9
107.95s$ ./gradlew assemble
Downloading https://repo.gradle.org/gradle/dist-snapshots/gradle-kotlin-dsl-4.1-20170728152452+0000-all.zip
......................................................................................
Unzipping /home/travis/.gradle/wrapper/dists/gradle-kotlin-dsl-4.1-20170728152452+0000-all/e0n85zynz8q7sl3rcq4r5ltay/gradle-kotlin-dsl-4.1-20170728152452+0000-all.zip to /home/travis/.gradle/wrapper/dists/gradle-kotlin-dsl-4.1-20170728152452+0000-all/e0n85zynz8q7sl3rcq4r5ltay
Set executable permissions for: /home/travis/.gradle/wrapper/dists/gradle-kotlin-dsl-4.1-20170728152452+0000-all/e0n85zynz8q7sl3rcq4r5ltay/gradle-4.1-20170728152452+0000/bin/gradle
Starting a Gradle Daemon (subsequent builds will be faster)
Generating JAR file 'gradle-api-4.1-20170728152452+0000.jar'
Download https://repo.gradle.org/gradle/repo/org/jetbrains/kotlin/kotlin-gradle-plugin/1.1.3-2/kotlin-gradle-plugin-1.1.3-2.pom
Download https://repo.gradle.org/gradle/repo/org/jetbrains/kotlin/kotlin-annotation-processing/1.1.3-2/kotlin-annotation-processing-1.1.3-2.pom
Download https://repo.gradle.org/gradle/repo/org/jetbrains/kotlin/kotlin-project/1.1.3-2/kotlin-project-1.1.3-2.pom
[...]

PR #493, opened 2017-09-12
Set JVMConfigurationKeys.JDK_HOME in KotlinCompiler for JDK 9 support

JLLeitschuh:
### Context
Setting the JDK_HOME configuration key explicitly for the Kotlin compiler
means that it can find the JDK 9 objects to compile against.

This may just be a stopgap fix and not the long term solution.

Related https://youtrack.jetbrains.com/issue/KT-20167
Related #454
Related #455

### Contributor Checklist
- [ ] Provide unit tests to verify logic (see #455 which should be merged after a release with this fix)

### DIFF
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
@@ -188,6 +188,12 @@ fun compilerConfigurationFor(messageCollector: MessageCollector, sourceFiles: It
     CompilerConfiguration().apply {
         addKotlinSourceRoots(sourceFiles.map { it.canonicalPath })
         addJvmClasspathRoots(PathUtil.getJdkClassesRootsFromCurrentJre())
+        /*
+         * Without this the compiler won't work with JDK 9.
+         * See the discussion here:
+         * https://youtrack.jetbrains.com/issue/KT-20167
+         */
+        put(JVMConfigurationKeys.JDK_HOME, File(System.getProperty("java.home")))
         put<MessageCollector>(CLIConfigurationKeys.MESSAGE_COLLECTOR_KEY, messageCollector)
     }
 


JLLeitschuh:
> Hi Rodrigo B. de Oliveira ,
The more safe way is to use CLI compiler (org.jetbrains.kotlin.cli.jvm.K2JVMCompiler) with -script-templates  compiler argument.
> 
> While we don't have strict compiler API, I think it's fine to use some of it internals but we can simply get bugs like this one.
> \- [Mikhail Zarechenskiy in Youtrack Comment](https://youtrack.jetbrains.com/issue/KT-20167#comment=27-2412729)

We may want to have a larger discussion about that but for now this is the simplest path to getting this to work.

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
```diff
@@ -188,6 +188,12 @@ fun compilerConfigurationFor(messageCollector: MessageCollector, sourceFiles: It
     CompilerConfiguration().apply {
         addKotlinSourceRoots(sourceFiles.map { it.canonicalPath })
         addJvmClasspathRoots(PathUtil.getJdkClassesRootsFromCurrentJre())
```
bamboo: Is this line still required if `JDK_HOME` is set?

JLLeitschuh: I have no idea. Personally, I'd rather not mess with it. I can run the tests locally and see if it works without it if you'd like.

bamboo:
There's a chance this won't be needed at all:

> we'll make fix to preserve behaviour for createForProduction  function
https://youtrack.jetbrains.com/issue/KT-20167#comment=27-2415450

Let's wait for the outcome of that discussion.

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
JLLeitschuh: I just ran `./gradlew check --parallel` without that line and pretty much all of the tests fail, so yes, it is still needed.

JLLeitschuh:
¯\_(ツ)_/¯

It would be nice to see this in a release sooner rather than later in my book and not have to wait on another Kotlin release to see this fixed. Fair enough though.

JLLeitschuh:
Given that Kotlin will be fixing this internally do you want to just wait for a release?

bamboo:
Let's move forward with the effort to reach JDK 9 compatibility and then revert this change once a Kotlin release with the proper fix is out.

bamboo:
Thanks, @JLLeitschuh!

Issue #424, opened 2017-07-20
Script plugins can be applied using the plugins {} DSL


bamboo:

eskatos:
This work will take place with the following steps:
1. make it work with the Groovy DSL in Gradle proper, including breaking changes for the Kotlin DSL to accomodate with
2. impact the Kotlin DSL to absorb the breaking changes without adding support for script plugins applied using the plugins block
3. implement script plugins applied using the plugins block in the Kotlin DSL

Depending on our findings along the way we may need to loop over changes in Gradle proper and the Kotlin DSL.

Step 1.  is currently being worked on at https://github.com/gradle/gradle/tree/pm/plugins-dsl/script-plugins

We made progress in allowing script plugins to be applied using the plugins DSL by introducing a new type of `PluginDependencySpec`, a `ScriptPluginDependencySpec` which is defined via the new script "path" dependency builder:

```kotlin
plugins {
  script("path/to/other.gradle")
  script("https://example.com/another.gradle")
}
```

By defining a new entry point in the DSL we make sure only attributes that make sense for script plugins can be specified.

And departing from the current `apply from: "path"` semantics, script paths are assumed to be relative to the requesting script, instead of the root project directory. We think this makes for clearer semantics, specially in the remote case and will allow remote script plugins to define their dependencies on other remote script plugins using relative paths.

At the time of writing, one issue remains for step 1: duplicate build operations events for script plugin application (from both `PluginManager` and `ScriptPlugin`).

This work also need to be rebased as https://github.com/gradle/gradle/pull/2600 has now been merged.

Reviewing for naming, responsibilities split, right abstractions and so on will be done next once the build operations are fixed.

oehme:
What will be the plugin ID of such a script? 

```
plugins {
  script("path/to/other.gradle")
  script("https://example.com/another.gradle")
}
```

bamboo:
The short answer is none.

In the long run it will be up for the user to assign an ID either via the `pluginManagement` or `plugins` DSL. For instance:

```kotlin
plugins {
  script("https://foo.bar/...") identifiedBy "foo" apply false
}

plugins.withId("foo") {
  // conditional logic depending on plugin being applied or not
}

// later...
if (someCondition) {
  apply { plugin("foo") }
}
```

bamboo:
There are two aspects to plugin identities:

 - providing an user identifiable handle for `plugins.withId(...)`, build operations, build scans, etc

[...]

PR #500, opened 2017-09-20
Move to the new PluginDependencySpec API

eskatos:
By using `org.gradle.plugin.dsl.PluginDependenciesSpec` from https://github.com/gradle/gradle/pull/2825.

Step for #424

### DIFF
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinBuildScriptCompiler.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinBuildScriptCompiler.kt
@@ -197,7 +197,7 @@ class KotlinBuildScriptCompiler(
 
         val (lineNumber, compiledScript) = compiledPluginsBlock
         val pluginsBlockClass = classFrom(compiledScript, baseScope.createChild("plugins"))
-        val pluginDependenciesSpec = pluginRequestCollector.createSpec(lineNumber)
+        val pluginDependenciesSpec = pluginRequestCollector.createPluginDependencySpec(lineNumber)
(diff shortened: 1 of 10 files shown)


provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinBuildScriptCompiler.kt
```diff
@@ -197,7 +197,7 @@ class KotlinBuildScriptCompiler(
 
         val (lineNumber, compiledScript) = compiledPluginsBlock
         val pluginsBlockClass = classFrom(compiledScript, baseScope.createChild("plugins"))
-        val pluginDependenciesSpec = pluginRequestCollector.createSpec(lineNumber)
+        val pluginDependenciesSpec = pluginRequestCollector.createPluginDependencySpec(lineNumber)
```
bamboo: :o: I understand this is coming from the Gradle API but shouldn't it be called `createPluginDependenciesSpec`?

eskatos: That's right! Let me fix this and I'll get back to you for another pass on this review.

eskatos: @bamboo fixed

JLLeitschuh:
Travis CI failed? Should we be concerned?

eskatos:
@JLLeitschuh we should not be, we validated the change before merging, but I agree this is definitely confusing. I opened #504 to track this down

JLLeitschuh:
Thank you!

PR #503, opened 2017-09-21
Expose groovy builder delegate

bamboo:


provider/src/test/kotlin/org/gradle/kotlin/dsl/GroovyInteroperabilityTest.kt
```diff
@@ -107,6 +112,7 @@ class GroovyInteroperabilityTest {
         val builderResult = delegate.withGroovyBuilder {
             val invokeResult = "withKeywordArguments"("string" to "42", "int" to 42)
             assertThat(invokeResult, sameInstance(expectedInvokeResult))
+            assertThat(this.delegate, sameInstance<Any>(delegate))
```
eskatos: ❓ Aren't `this.delegate` and `delegate` the same reference?

bamboo: They are not. `delegate` refers to the local variable. The rule is that local variables have precedence over implicit receiver members. But declaring an explaining variable in the outer scope makes sense and I'll do that.

eskatos: > The rule is that local variables have precedence over implicit receiver members.

I keep forgetting this, not obvious to me, I need to stick this in my head

> But declaring an explaining variable in the outer scope makes sense and I'll do that.

👍

bamboo: Yeah, I remember this rule as the only disappointment I had with Kotlin so far.

JLLeitschuh:
Can you provide an example of how this would be used in practice? The tests you have don't really show a practical "this is how this would be used" use case.
Maybe an update to the documentation regarding how this should be used?

bamboo:
@JLLeitschuh it seems you missed the commit comment.

A proper sample will come in a future update.

JLLeitschuh:
I don't normally read the commit comments. But now that I have I see the example. Cool! Can you include the commit comment in the description of the PR in the future?

provider/src/test/kotlin/org/gradle/kotlin/dsl/GroovyInteroperabilityTest.kt
bamboo: @eskatos Done.

bamboo:
@JLLeitschuh I'll make an effort.

brettwooldridge:
@bamboo @JLLeitschuh What am I missing?  I have:
```kotlin
"beforeDeployment" {
   if (signing.isRequired)
      signing.signPom(delegate as MavenDeployment)
}
```
in my ``build.gradle.kts`` file, but I am getting the following error:
```
e: /Users/brettw/Documents/dev/jnb-ping/build.gradle.kts:88:38: Unresolved reference: delegate
```

JLLeitschuh:
I don't think this feature is in the current release. You will probably need to either use the bleeding edge build or wait for this feature to be in `4.3`.

brettwooldridge:
Is there an alternative pattern for jar signing that works with the current version?

bamboo:
@brettwooldridge I don't see `withGroovyBuilder { ... }` in your sample code. `delegate` is only valid within the scope of `withGroovyBuilder`. For an example of `withGroovyBuilder` please see [samples/maven-plugin/build.gradle.kts#L20](https://github.com/gradle/kotlin-dsl/blob/develop/samples/maven-plugin/build.gradle.kts#L20).

brettwooldridge:
@JLLeitschuh Actually, I just checked, and I'm running ``gradle-kotlin-dsl-4.3-20171004164220+0000``.  The above ``build.gradle.kts`` snippet came from the reference [in the v0.11.1 release announcement](https://github.com/gradle/kotlin-dsl/releases).

brettwooldridge:
@bamboo Thanks.  Moving it into the ``withGroovyBuilder`` scope worked.

PR #507, opened 2017-09-22
Fix typo to use correct system property value to execute Kotlin compi…

mkobit:
…ler in process

see https://discuss.kotlinlang.org/t/solved-disable-kotlin-compiler-process-keepalive/2622

### Contributor Checklist
- [ ] ~~Provide integration tests to verify changes from a user perspective~~
- [ ] ~~Provide unit tests to verify logic~~

### Notes

- `./gradlew :provider:check --info --stacktrace --console=plain --max-workers=1 --no-daemon -Dkotlin.compiler.execution.strategy="in-process" -Dkotlin.colors.enabled=false` succeeds locally

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ matrix:
           packages:
             - oracle-java9-installer # Forces use of even newer JDK 9 build
 
-script: ./gradlew :provider:check --info --stacktrace --console=plain --max-workers=1 --no-daemon -Dkotlin.compiler.execution.strategy=in-proces -Dkotlin.colors.enabled=false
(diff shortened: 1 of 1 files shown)


eskatos:
In Kotlin 1.1.50: https://github.com/JetBrains/kotlin/blob/v1.1.50/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/compilerRunner/GradleKotlinCompilerRunner.kt#L43

bamboo:
Good catch! 👍

Issue #465, opened 2017-08-16
Provide a way to configure default kotlin plugins/artifacts version 


ligee:
Now the `kotlin(module)` functions for `dependencies` and `plugins` block either accept an explicit version of the module or take the version of the embedded kotlin distribution.

It would be helpful to provide a way to customize the default version, e.g. via the `org.gradle.kotlin.dsl.default-kotlin-version` (system) property.

JLLeitschuh:
The current default is the version of kotlin that is being used to compile your Gradle Kotlin DSL files (I think). Is there a way to decouple these? Does it make sense to decouple these?

Can you compile your Gradle Kotlin DSL files with another version of Kotlin?

gildor:
I would repeat my comment from Gradle channel in Kotlin Slack.

> It looks strange for me that kotlin-dsl provides shortcut that specific only for Kotlin project (`kotlin()` or `kotlinModule()` before)
> 
> `kotlin(“…”)` method in kotlin-dsl is really convenient but forces you to use the wrong approach, because your project implicitly relates now on a version of Kotlin bundled to kotlin-dsl this approach doesn’t look like a recommended way to configure your project.
>Of course, you can specify version explicitly, but I think most developers don’t do that.
> 
> if you use kotlin-gradle-plugin with new plugins DSL you can omit version as well and you always want to use the same version of the plugin with the same version of other kotlin dependencies. 
> Maybe better to move kotlin(“module”) extension to kotlin-gradle-plugin to allow use it. Looks like much better way to help configure project, actually it’s just additional helper function in plugin-specific DSL.
> 
> I thought that Gradle is not biased to any specific project type and I think it’s a good thing and kotlin-dsl should follow this strategy.
> 
> P.S. I very like idea of helper methods for project to reduce copy-paste of dependency ids and boiler plate extraction of dependency versions.
> I would really want such thing in Android plugin for support-library dependencies. But it should be a part of plugin, not Gradle or kotlin-dsl API

My proposal to remove `kotlin()` method from kotlin-dsl and move it to Kotlin plugin.

bamboo:
Thanks for the proposal, @gildor. 

I've provided the rationale for the `kotlin` / `kotlinModule` dependency notation and `embeddedKotlinVersion` property in the original issue: #94.

jmfayard:
Here is a proposal I made in slack

```
plugins {
    kotlin(module = "jvm", version = "1.1.4-3")
}
dependencies {
[...]

PR #511, opened 2017-09-27
DependencySpec.kotlin extension no more defaults to embeddedKotlinVersion

eskatos:
See #465, especially https://github.com/gradle/kotlin-dsl/issues/465#issuecomment-332543198


buildSrc/src/main/kotlin/codegen/GenerateKotlinDependencyExtensions.kt
```diff
@@ -74,10 +74,10 @@ fun RepositoryHandler.gradleScriptKotlin(): ArtifactRepository =
  * Builds the dependency notation for the named Kotlin [module] at the given [version].
  *
  * @param module simple name of the Kotlin module, for example "reflect".
- * @param version optional desired version, null implies [embeddedKotlinVersion].
+ * @param version optional desired version, unspecified if null.
  */
 fun DependencyHandler.kotlin(module: String, version: String? = null): Any =
-    "org.jetbrains.kotlin:kotlin-${'$'}module:${'$'}{version ?: embeddedKotlinVersion}"
+    "org.jetbrains.kotlin:kotlin-${'$'}module${'$'}{if(version == null) "" else ":${'$'}version"}"
```
JLLeitschuh: What's with the wonky syntax here?

eskatos: Hi Jonathan,

This form of [string template escaping](https://kotlinlang.org/docs/reference/basic-types.html#string-templates) is necessary here because this is code generating code that includes a string template.

Were you referring to something else?

JLLeitschuh: Oh, okay, I kind of understand.
Why are you using `${'$'}` instead of simply doing `\$`?

eskatos: Both links ☝️  explain that you must use this escaping style when using raw strings (aka. multi-line strings), which is the case here.

JLLeitschuh: Ah, fair enough. That makes much more sense.

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/KotlinBuildScriptModelIntegrationTest.kt
```diff
@@ -186,7 +186,7 @@ class KotlinBuildScriptModelIntegrationTest : AbstractIntegrationTest() {
             rootProjectScript = "",
             subProjectScript ="""
                 buildscript {
-                    dependencies { classpath(kotlin("gradle-plugin")) }
+                    dependencies { classpath(kotlin("gradle-plugin", "$embeddedKotlinVersion")) }
```
bamboo: :x: We should introduce a `DependencyHandler.embeddedKotlin(module: String)` dependency builder to replace the old usages.

samples/hello-js/build.gradle.kts
```diff
@@ -2,7 +2,7 @@ import org.jetbrains.kotlin.gradle.tasks.Kotlin2JsCompile
 
 buildscript {
     dependencies {
-        classpath(kotlin("gradle-plugin"))
+        classpath(kotlin("gradle-plugin", "1.1.50"))
```
bamboo: :x: We should also introduce test cases to make sure we don't forget to update the version whenever we bump the embedded Kotlin version.

buildSrc/src/main/kotlin/codegen/GenerateKotlinDependencyExtensions.kt
bamboo: :0: I would express this using `?let` and null-coalescing `version?.let { ":${'$'}version" } ?: ""`

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/KotlinBuildScriptModelIntegrationTest.kt
eskatos: Done

samples/hello-js/build.gradle.kts
eskatos: Done

buildSrc/src/main/kotlin/codegen/GenerateKotlinDependencyExtensions.kt
eskatos: 👍 Done

eskatos:
@bamboo changes pushed, could you have another look?

Issue #512, opened 2017-09-27
Extra Defaults change in commit c21eba6d looses type information


sic-bre:
Commit c21eba6d does more than "Adjust code to nullability changes in the Gradle API".  Previously when using extra defaults to access gradle properties the information of the type of the property was maintained.  Now the type of the property comes back as Any and I've a had to add casts in contexts where the type is important.  This change was made in response to in Gradle 4.2 the PropertyState setter having the @Nullable annotation applied.

### Expected Behavior

Instead I suggest that the type value passed in the Kotlin Extension should have been changed from T to T? rather than Any.  Allowing a nullable value signature but preserving types for code that uses the extra property types.

### Current Behavior

In ExtraPropertiesExtensions.kt value was changed from T to Any on setValue to "accommodate a change in nullablity".  However this changes looses the type information that was previously passed.

operator fun ExtraPropertiesExtension.setValue(receiver: Any?, property: KProperty<*>, value: **Any**) =  set(property.name, value)

Several other places in the file the type was changed from T to **T: Any**.

### Context

I have created a plugin that depended on the extra property extensions passing correct type information.  This worked as expected in Gradle 4.1 but with Gradle 4.2 latest type signature change this type information is getting removed and replaced with Any.

### Steps to Reproduce (for bugs)
Functional change in type signatures.

### Your Environment
Behavior changed between Gradle 4.1 and Gradle 4.2.  The commit that made the change was made https://github.com/gradle/kotlin-dsl/commit/21eba6dc6e8d3a672c530e680481a1b367ef031f

JLLeitschuh:
I think this is the change being referred to by @sic-bre:
https://github.com/gradle/kotlin-dsl/commit/21eba6dc6e8d3a672c530e680481a1b367ef031f#diff-e6df212bc9b2bc8b63741a3b7078ff5bL37

bamboo:
Thanks for the report, @sic-bre, we'll look into it.

eskatos:
One of the underlying issue is that `ExtraPropertiesExtension` in the Gradle API does not declare nullability properly.

- [`Object get(String name)`](https://github.com/gradle/gradle/blob/b0032a8a355379e5868b22d660f8be5cf857168e/subprojects/core-api/src/main/java/org/gradle/api/plugins/ExtraPropertiesExtension.java#L121) method should be `@Nullable`
[...]

PR #516, opened 2017-09-29
Restore extra properties type info & allow nullability

eskatos:
Optional extra properties can now be expressed via nullability of the property type.

```kotlin
val name by extra { null } // creates null extra property

val name: String by extra  // throws NPE on property usage
val name: String? by extra // works as expected
```

This PR also enhances the `extra-properties` sample to demonstrate this.

See #512

### DIFF
--- a/samples/extra-properties/build.gradle.kts
+++ b/samples/extra-properties/build.gradle.kts
@@ -2,28 +2,19 @@ val myTask = task("myTask") {
 
     val foo by extra { 42 }
     val bar by extra<Int?>(null)
-    val bazar: Int? by extra { null }
 
     doLast {
         println("Extra foo property value: $foo")
         println("Optional extra bar property value: $bar")
-        println("Optional extra bazar property value: $bazar")
     }
 }
 
 val foo: Int by myTask.extra
 val bar: Int? by myTask.extra
-val bazar: Int by myTask.extra
 
 afterEvaluate {
     println("myTask.foo = $foo")
     println("myTask.bar = $bar")
-    try {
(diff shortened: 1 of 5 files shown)


samples/extra-properties/build.gradle.kts
```diff
@@ -1,15 +1,29 @@
...
+val bar: Int? by myTask.extra
+val bazar: Int by myTask.extra
+
 afterEvaluate {
     println("myTask.foo = $foo")
+    println("myTask.bar = $bar")
+    try {
+        println("myTask.bazar = $bazar")
+        require(false, { "Should not happen as `bazar`, requested as a Int is effectively null" })
+    } catch (ex: NullPointerException) {
```
bamboo: :o: I'm ambivalent about having the `try / catch` in the sample. I understand the motivation for demonstrating what happens when the declaration sites disagree on the nullability but it looks a little overwhelming for a sample.

eskatos: Yeah, thought the same.
I'll move this particular case as an integration test instead.

sic-bre:
Which version of Gradle or the kotlin-dsl project should I use to test these changes against my project?

bamboo:
Hi @sic-bre, these changes will be officially available in the upcoming Gradle 4.3 release but you can test them Today with the latest Gradle Kotlin DSL distribution snapshot by updating your Gradle wrapper in the following manner:

    ./gradlew --gradle-distribution-url https://repo.gradle.org/gradle/dist-snapshots/gradle-kotlin-dsl-4.3-20171003164627+0000-all.zip

PR #536, opened 2017-10-10
Fix broken android sample

eskatos:


samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/SamplesSmokeTest.kt
```diff
@@ -31,7 +31,9 @@ class SamplesSmokeTest(
 
     @Before
     fun populateProjectRootWithSample() {
-        assumeThat(sampleName, not(containsString("android")))
+        if (sampleName.contains("android")) {
+            assumeTrue(System.getenv().containsKey("ANDROID_HOME"))
+        }
```
JLLeitschuh: Can you leave a bit of an explanation why this test is disabled in this case? (Preferably in a comment in the code instead of in a PR comment).

PR #540, opened 2017-10-12
Account for PropertyState deprecation in favor of Property

eskatos:
And `project.property()` in favor of `project.objects.property()`

### DIFF
--- a/samples/provider-properties/build.gradle.kts
+++ b/samples/provider-properties/build.gradle.kts
@@ -29,7 +29,7 @@ open class GreetingPlugin : Plugin<Project> {
 open class GreetingPluginExtension(project: Project) {
 
     private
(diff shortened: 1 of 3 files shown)


samples/provider-properties/build.gradle.kts
```diff
@@ -29,7 +29,7 @@ open class GreetingPlugin : Plugin<Project> {
 open class GreetingPluginExtension(project: Project) {
 
     private
-    val messageState = project.property<String>()
+    val messageState = project.objects.property<String>()
```
bamboo: :x: Take the opportunity and rename to `messageProperty`?

eskatos: Done

PR #541, opened 2017-10-12
Support `Settings` as a script target

eskatos:
This PR enables the use of settings scripts using the Kotlin DSL, aka. `settings.gradle.kts`


provider/src/main/kotlin/org/gradle/kotlin/dsl/SettingsExtensions.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import org.gradle.api.initialization.Settings
+
+/**
+ * Adds the given projects to the build.
+ *
+ * @param projectPaths the projects to add
+ * @see [Settings.include]
+ */
+fun Settings.include(vararg projectPaths: String) {
```
bamboo: ⭕️  I assume we're adding this here because it would be a breaking change to do it on the Gradle side?

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/CachingKotlinCompiler.kt
```diff
@@ -101,16 +100,17 @@ class CachingKotlinCompiler(
 
     data class CompiledPluginsBlock(val lineNumber: Int, val compiledScript: CompiledScript)
 
-    fun compileBuildScript(
+    fun compileGradleScript(
         scriptPath: String,
         script: String,
+        scriptTemplate: KClass<out Any>,
```
bamboo: ⭕️  I think the `scriptTemplate` deserves to be the first in the argument list like it is elsewhere.

provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/KotlinBuildScriptDependenciesResolver.kt
```diff
@@ -219,7 +219,9 @@ class KotlinBuildScriptDependencies(
 internal
 fun projectRootOf(scriptFile: File, importedProjectRoot: File): File {
 
-    fun isProjectRoot(dir: File) = File(dir, "settings.gradle").isFile
+    // TODO:pm remove hardcoded reference to settings.gradle
```
bamboo: :o: Maybe use `BuildLayoutFactory` as a service here?

provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/ProjectSchemaAccessorsIntegrationTest.kt
```diff
@@ -126,8 +126,8 @@ class ProjectSchemaAccessorsIntegrationTest : AbstractIntegrationTest() {
     @Test
     fun `can access configurations registered by declared plugins via jit accessor`() {
 
-        withFile("settings.gradle", """
-            include "a", "b", "c"
+        withFile("settings.gradle.kts", """
```
bamboo: ⭕️  We should leave at least one test checking the combination `settings.gradle` / `build.gradle.kts`.

provider/src/test/kotlin/org/gradle/kotlin/dsl/resolver/ProjectRootOfTest.kt
```diff
@@ -6,17 +6,27 @@ import org.hamcrest.CoreMatchers.equalTo
...
+import org.junit.runners.Parameterized
 
-class ProjectRootOfTest : FolderBasedTest() {
+@RunWith(Parameterized::class)
+class ProjectRootOfTest(private val settingsFileName: String) : FolderBasedTest() {
+
+    companion object {
+        @Parameterized.Parameters(name = "{0}")
+        @JvmStatic
+        fun testCases(): Iterable<Array<Any>> =
```
bamboo: :o: The return type is unnecessary here but if you must declare it, why not `Iterable<Array<String>>`?

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptPluginTarget.kt
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+
+internal
+fun unsupportedTarget(target: Any): Nothing =
+    throw IllegalArgumentException("Unsupported target ${target::class.qualifiedName}: $target")
+
+
+internal
+interface KotlinScriptPluginTarget<T : Any> {
```
bamboo: :x: Wouldn't this be simpler with a data class?

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptPlugin.kt
```diff
@@ -37,14 +36,18 @@ class KotlinScriptPlugin(
...
     override fun apply(target: Any) {
         logger.debug("Applying Kotlin script to {}", target)
-        if (target !is Project) {
-            throw IllegalArgumentException("target $target was not a Project as expected")
-        }
-        target.run {
-            afterEvaluate {
-                plugins.apply(KotlinScriptBasePlugin::class.java)
-            }
+        when (target) {
```
bamboo: :x: We have already taken a decision based on the target before, could we leverage that somehow to avoid taking this decision again here?

plugins/src/test/kotlin/org/gradle/kotlin/dsl/plugins/AbstractPluginTest.kt
```diff
@@ -15,11 +15,11 @@ open class AbstractPluginTest : AbstractIntegrationTest() {
     fun setUpTestPluginRepository() {
         val testRepository = normaliseFileSeparators(absolutePathOf("build/repository"))
         val futureVersion = loadTestProperties()["version"]
-        withFile("settings.gradle", """
+        withFile("settings.gradle.kts", """
             pluginManagement {
                 repositories {
-                    maven { url = uri("$testRepository") }
-                    maven { url = uri("https://jcenter.bintray.com/") }
+                    maven { setUrl("$testRepository") }
```
bamboo: :o: Let's not forget to add follow up issues to recover the nice DSL.

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinBuildScriptCompiler.kt
```diff
@@ -68,70 +69,77 @@ class KotlinBuildScriptCompiler(
...
         when {
             topLevelScript -> compileTopLevelScript()
             else           -> compileScriptPlugin()
         }
 
-    fun compileForClassPath(): (Project) -> Unit = { target ->
-        ignoringErrors { executeBuildscriptBlockOn(target) }
-        ignoringErrors { prepareTargetClassLoaderScopeOf(target) }
-        ignoringErrors { executeScriptBodyOn(target) }
+    fun asKotlinScriptPluginTarget(configuration: (KotlinScriptPluginTarget<*>) -> Unit): (Any) -> Unit = { anyTarget ->
```
bamboo: 👍 Nice move introducing `KotlinScriptPluginTarget`.

provider/src/test/kotlin/org/gradle/kotlin/dsl/resolver/ProjectRootOfTest.kt
JLLeitschuh: Because of Junit 4 weirdness, I don't think that will work. I've dealt with Junit 4's Parameterized tests and this can be devilishly tricky to get right.

provider/src/main/kotlin/org/gradle/kotlin/dsl/SettingsExtensions.kt
JLLeitschuh: Please excuse my lack of knowledge of the java language, but wouldn't changing:
```java
void include(String[] projectPaths);
```
to
```java
void include(String... projectPaths);
```
still allow you to pass an array into the `include` method? Maybe this is just a misunderstanding of Java method signatures.

JLLeitschuh:
I'm not familiar enough with the code to know if you have this already, but make sure that you have the sam-with-conversion plugin working correctly in these files.
It might make sense to add a test that makes sure that it works with either `pluginManagement` or `includeBuild`.

Other things to ensure that work correctly:

In `settings.gradle`, I can do something like this:
```gradle
File toInclude = file("../someOtherProjectToInclude")
if (toInclude.exists()) {
    includeBuild(toInclude)
}
```
The problem is that the `file` is defined on `Plugin` which isn't in the type hierarchy for the `Settings` type.

I'm uncertain that you would be able to do this in your `settings.gradle.kts` file the way you have this now:
```kotlin
file("../someOtherProjectToInclude")
```

For `PluginResolveDetails` do you want to also offer an overload for `useModule` that allows you to use named parameters? Eg:
```kotlin
useModule(group = "someGroup", name = "someName", version = "someVersion")
```

Also, the `BuildCacheConfiguration` type could use some helpful overloads to use kotlins reified types.

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/CachingKotlinCompiler.kt
eskatos: Done

provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/KotlinBuildScriptDependenciesResolver.kt
eskatos: I think we should not use an internal Gradle service here. And this code runs in IntelliJ.
See my comment on 3035e37457f2fb0cf72fe8efe8c4419e5319188d

provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/ProjectSchemaAccessorsIntegrationTest.kt
eskatos: Done

provider/src/test/kotlin/org/gradle/kotlin/dsl/resolver/ProjectRootOfTest.kt
eskatos: Done.
It does work without the explicit return type, see https://github.com/gradle/kotlin-dsl/pull/541/files#diff-c27c161d1436684b5a027b2b908101e0R18

plugins/src/test/kotlin/org/gradle/kotlin/dsl/plugins/AbstractPluginTest.kt
eskatos: https://github.com/gradle/kotlin-dsl/issues/543

provider/src/main/kotlin/org/gradle/kotlin/dsl/SettingsExtensions.kt
eskatos: I removed these premature extensions and postponed this aspect as #543

eskatos:
Thank you for the suggestions @JLLeitschuh, nice DSL in `settings.gradle.kts` will be done in #543.

eskatos:
As for the `sam-with-receiver` compiler plugin, it is already covered see https://github.com/gradle/kotlin-dsl/blob/0dbba5f0729cb82b3d5b196cf1c2a187ea146945/plugins/src/test/kotlin/org/gradle/kotlin/dsl/plugins/AbstractPluginTest.kt#L18-L32

JLLeitschuh:
Is `file` supported currently the way you have this feature designed? That was one that I was really unsure about.

eskatos:
Nope, it should come with #543 (_"Implicit ones from `org.gradle.api.Script` should also be considered._")

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptPlugin.kt
eskatos: Actually this is the "earliest" decision point. The second one you refer too happen inside `script: (Any) -> Unit` here. We don't have the target at the point where we instantiate `KotlinScriptPlugin`.

JLLeitschuh:
Thanks, makes more sense now.

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptPlugin.kt
eskatos: I moved this responsibility to `KotlinScriptPluginTarget`, much cleaner!

eskatos:
@bamboo, could you have another look?

eskatos:
Thanks for your review @bamboo, I addressed your comments and added your suggested changes to this PR. Could you have another look?

Issue #550, opened 2017-10-17
Update all samples to use settings.gradle.kts


eskatos:

JLLeitschuh:
Do you want to wait until the kotlin plugin to be released as stable?

eskatos:
Not necessarily, the Kotlin plugin update is about IntelliJ editing support only and we want to move forward with the samples on our `develop` branch.

Without Kotlin IntelliJ plugin 1.1.60 you can still edit the `settings.gradle.kts` files but they are "all red" because their base class and classpath are wrong (same as a `Project` script plugin).

And 1.1.60 is just around the corner anyway.

eskatos:
Moreover, this will exercise the support for `settings.gradle.kts` earlier as we integration test all samples. In practice this may be done in several incremental steps following progress on #56.

JLLeitschuh:
Sounds good. Given that master will keep the samples that use `settings.gradle` this makes sense.

eskatos:
Closing as delivered in #555, #570, #571 and #572

PR #555, opened 2017-10-18
Samples and build use settings.gradle.kts

eskatos:
See #550


JLLeitschuh:
Android studio requires that a `build.gradle` file exists. It doesn't have the same requirement for `settings.gradle` to exist correct?

JLLeitschuh:
Another question, when developing composite build projects the project that you are targeting in an `includeBuild` call requires that targeted project have a `settings.gradle` file (you get an exception if you don't).

Two questions:
 1. Have you tested that a composite build project with a `settings.gradle.kts` file and without a `settings.gradle` file works correctly?
 2. Has the exception mentioning that you must have a `settings.gradle` file been updated to state that you need either a `settings.gradle` or a `settings.gradle.kts` file?

bamboo:
> Android studio requires that a build.gradle file exists. It doesn't have the same requirement for settings.gradle to exist correct?

That's a good question.

JLLeitschuh:
I've opened an issue with TravisCI about all of the JDK 9 builds failing:
https://github.com/travis-ci/travis-ci/issues/8608

eskatos:
> Android studio requires that a build.gradle file exists. It doesn't have the same requirement for settings.gradle to exist correct?

Correct. I validated that project import and usage in Android Studio is working.

> Another question, when developing composite build projects the project that you are targeting in an includeBuild call requires that targeted project have a settings.gradle file (you get an exception if you don't).
>
> Two questions:
> - Have you tested that a composite build project with a settings.gradle.kts file and without a settings.gradle file works correctly?
> - Has the exception mentioning that you must have a settings.gradle file been updated to state that you need either a settings.gradle or a settings.gradle.kts file?

Yes, this has been handled upstream in Gradle.

Issue #558, opened 2017-10-19
Ship `kotlin-stdlib-jre8`


bamboo:
It seems it should be safe to ship it as long as we don't directly use it but only expose the library to build scripts:

> There are now separate versions of the standard library supporting the new JDK APIs added in Java 7 and 8. If you need access to the new APIs, use kotlin-stdlib-jre7 and kotlin-stdlib-jre8 maven artifacts instead of the standard kotlin-stdlib. These artifacts are tiny extensions on top of kotlin-stdlib and they bring it to your project as a transitive dependency.
https://kotlinlang.org/docs/reference/whatsnew11.html#java-8-standard-library-support

ilya-g:
Note that in Kotlin 1.2 these extension artifacts will be deprecated, so you might want to replace them with `-jdk7` and `-jdk8` versions then.

bamboo:
Thanks for the heads-up, @ilya-g.

PR #559, opened 2017-10-19
Expose `kotlin-stdlib-jre8` to build scripts

bamboo:
See #558

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -44,7 +44,7 @@ tasks.withType(KotlinCompile) {
 dependencies {
     compile gradleApi()
     compile "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
-    compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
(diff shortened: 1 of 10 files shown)


JLLeitschuh:
Can you make sure that this is rebased on top of the `develop` branch so that JDK 9 travis fix is in there??

JLLeitschuh:
So you ran the Open JDK build twice (once as a push check and once as a PR check).

One of the openJDK build tests fails with what looks like a compile error but Its hard to tell.
I'm confused why it fails once but not the other time.
https://travis-ci.org/gradle/kotlin-dsl/jobs/290168068

bamboo:
@eskatos done. Looks like a Kotlin bug to me.

eskatos:
Indeed. Mind finding/opening an issue upstream?

JLLeitschuh:
> Retrieving groups by name is not supported on this platform.

That is really strange.

bamboo:
Follow-up Kotlin YT issue: https://youtrack.jetbrains.com/issue/KT-20865

PR #567, opened 2017-10-24
More samples that demonstrate the use of settings.gradle.ts

eskatos:
See #550


bamboo:
Hey @eskatos, could you post each sample as a separate PR please? That will allow some samples to hit `develop` faster.

By the way, if we really need a git repository as part of one of the samples, I strongly feel it should be generated from sources by a Gradle task.

eskatos:
@bamboo, makes sense, will do, closing this PR

Issue #221, opened 2017-01-12
Generate extensions taking a `KClass` for methods in the Gradle API taking a `java.lang.Class` 


bamboo:
So we can write `project.container(Foo::class)` instead of `project.container(Foo::class.java)`. 

Investigate the possibility of making this a core Kotlin feature (or contribute one via a compiler plugin) just like SAM conversions.

JLLeitschuh:
@eskatos What happened to the thought of making this a compiler plugin?

eskatos:
Closing as delivered in #914

PR #570, opened 2017-10-24
Build-cache sample

eskatos:
See #550


samples/build-cache/settings.gradle.kts
```diff
@@ -0,0 +1,21 @@
...
+    local {
+        isEnabled = true
+    }
+    remote(DirectoryBuildCache::class.java) {
+        setDirectory("path/to/some/local-cache")
+        isEnabled = true
+        isPush = false
+        targetSizeInMB = 1024
+    }
+    remote(HttpBuildCache::class.java) {
```
JLLeitschuh: This API could use some reified type love.

Not a part of this PR though.

bamboo: Indeed.

bamboo: That will be part of #221.

PR #571, opened 2017-10-24
Demonstrate dependency substitution in composite build sample

eskatos:
See #550


eskatos:
`README` added, could you have another look @bamboo?

Issue #543, opened 2017-10-12
Settings scripts DSL parity with Groovy


eskatos:
The DSL available in `settings.gradle.kts` file after #541 is quite rough in some places.

This issue is about "restoring" a nice DSL in settings scripts when using the Kotlin DSL.

From `org.gradle.api.initialization.Settings`:
- [x] `pluginManagement {}`
- [x] `include(..)` & `includeFlat(..)`
- [x] `includeBuild(..) {}`
- [x] `buildCache {}`
- [x] `sourceControl {}`

And
- [x] implicit ones from `org.gradle.api.Script`

eskatos:
Delivered via gradle/gradle#3187, #579, #589 and gradle/gradle#3347.

Subsequent refinements should be addressed as separate issues, closing.

PR #579, opened 2017-10-27
All Kotlin scripts implement org.gradle.kotlin.dsl.provider.ScriptApi

eskatos:
Introduce `org.gradle.kotlin.dsl.provider.ScriptApi` as a Kotlin flavored equivalent of `org.gradle.api.Script` that defines a common API to be supported by all script templates.

It is not implemented directly by script templates to overcome ambiguous conflicts, kotlin/java interop, platform types nullability etc.. A reflection based unit test validates that all script templates contains compatible members.

The validation handles a range of selected use cases supporting the introduced api.

This work is part of #543

### DIFF
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/ScriptApi.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/ScriptApi.kt
@@ -45,6 +45,7 @@ import java.net.URI
  * Members here must not use default parameters values.
  * Documentation should go to the actual implementations so that it shows up in IDEs properly.
  */
+@Suppress("unused")
 internal
 interface ScriptApi {
 
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/ScriptApi.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/ScriptApi.kt
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.gradle.kotlin.dsl.provider
+
+import org.gradle.api.PathValidation
+import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.file.ConfigurableFileTree
+import org.gradle.api.file.CopySpec
+import org.gradle.api.file.DeleteSpec
+import org.gradle.api.file.FileTree
+import org.gradle.api.logging.Logger
+import org.gradle.api.logging.LoggingManager
+import org.gradle.api.resources.ResourceHandler
+import org.gradle.api.tasks.WorkResult
+import org.gradle.process.ExecResult
+import org.gradle.process.ExecSpec
+import org.gradle.process.JavaExecSpec
+
+import java.io.File
+import java.net.URI
+
+/**
+ * Base contract for all Gradle Kotlin DSL scripts.
+ *
+ * This is the Kotlin flavored equivalent of [org.gradle.api.Script].
+ *
+ * It is not implemented directly by script templates to overcome ambiguous conflicts and Kotlin language
+ * limitations. See each script template for actual implementations.
+ *
+ * `ScriptApiTest` validates that each script template provide compatible methods/properties.
+ *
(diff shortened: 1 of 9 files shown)


provider/src/test/kotlin/org/gradle/kotlin/dsl/provider/ScriptApiTest.kt
```diff
@@ -0,0 +1,165 @@
...
+private
+fun KType.hasCompatibleTypeArguments(api: KType) =
+    arguments.size == api.arguments.size && (0..(arguments.size - 1)).all { idx ->
+        arguments[idx].type!!.isTypeArgumentCompatibleWith(api.arguments[idx].type!!)
+    }
+
+
+private
+fun KType.isTypeArgumentCompatibleWith(api: KType) =
+    withNullability(false) == api
```
eskatos: This is very naive but works for currently covered use cases.
I couldn't find a way to distinguish `Any?` from `Any!` nor investigated querying jsr-305.

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/ScriptApi.kt
```diff
@@ -0,0 +1,79 @@
...
+    fun copySpec(): CopySpec
+    fun copySpec(configuration: CopySpec.() -> Unit): CopySpec
+
+    fun mkdir(path: Any): File
+
+    fun delete(vararg paths: Any): Boolean
+    fun delete(configuration: DeleteSpec.() -> Unit): WorkResult
+
+    fun exec(configuration: ExecSpec.() -> Unit): ExecResult
+    fun javaexec(configuration: JavaExecSpec.() -> Unit): ExecResult
```
bamboo: :o: Should we take the opportunity to "fix" that name and use `javaExec` instead?

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
```diff
@@ -502,6 +502,37 @@ class GradleKotlinDslIntegrationTest : AbstractIntegrationTest() {
...
+    @Test
+    fun `scripts can use the gradle script api`() {
+
+        fun usageFor(target: String) = """
+
+            logger.error("Error logging from $target")
+            require(logging is LoggingManager, { "logging" })
+            require(resources is ResourceHandler, { "resources" })
+
+            require(relativePath("settings.gradle.kts") == "settings.gradle.kts", { "relativePath(path)" })
```
bamboo: :x: I think it's important for at least one test to capture the _paths are relative to the root project dir_ semantics, did I miss one?

provider/src/test/kotlin/org/gradle/kotlin/dsl/provider/ScriptApiTest.kt
```diff
@@ -0,0 +1,165 @@
...
+import org.hamcrest.CoreMatchers.equalTo
+import org.junit.Assert.assertThat
+import org.junit.Test
+
+
+class ScriptApiTest {
+
+    @Test
+    fun `build script template implements script api`() =
+        assertThat(
```
bamboo: :o: I wonder if it wouldn't be clearer and require a little code to first convert the candidate members to some canonical representation (data class or string) and then assert something like `assertThat(candidateApiMembers, hasItems(apiMembers))` where `candidateApiMembers` and `apiMembers` are of type `List<<canonical representation>>`.

samples/build-cache/settings.gradle.kts
```diff
@@ -3,13 +3,13 @@ buildCache {
         isEnabled = true
     }
     remote(DirectoryBuildCache::class.java) {
-        setDirectory("path/to/some/local-cache")
+        directory = file("path/to/some/local-cache")
```
bamboo: 👍 Nice!

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/ScriptApi.kt
eskatos: I don't think we should. This would be disturbing to users porting from Groovy, reading online materiel etc... with no real benefit.

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
eskatos: I made a slight change to this assertion to capture that semantic for `settings.gradle.kts` scripts.

It doesn't really capture it for the build script case, but I don't think we need more coverage here, this is already largely covered upstream.

provider/src/test/kotlin/org/gradle/kotlin/dsl/provider/ScriptApiTest.kt
eskatos: I'd say that it looks tempting but the reflect api is already somehow canonical. There are a lot of cases to cover and that would probably be quite a lot of code.

The existing assertion outputs the missing members nicely when there are any.

bamboo:
Could you fix the unused warnings in `ScriptApi` before merging though?

PR #589, opened 2017-11-08
Leverage upstream changes to settings pluginManagement.repositories

eskatos:
- use named repositories where applicable
- remove workarounds to make plugin resolution across composite builds work
- update `Configuring-Plugins.md` doc

### DIFF
--- a/plugins/src/test/kotlin/org/gradle/kotlin/dsl/plugins/AbstractPluginTest.kt
+++ b/plugins/src/test/kotlin/org/gradle/kotlin/dsl/plugins/AbstractPluginTest.kt
@@ -18,7 +18,7 @@ open class AbstractPluginTest : AbstractIntegrationTest() {
         withSettings("""
             pluginManagement {
                 repositories {
-                    maven { setUrl("$testRepository") }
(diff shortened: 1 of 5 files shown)


samples/hello-js/settings.gradle.kts
```diff
@@ -1,10 +1,4 @@
 pluginManagement {
-    repositories {
```
bamboo: Yay!

plugins/src/test/kotlin/org/gradle/kotlin/dsl/plugins/AbstractPluginTest.kt
```diff
@@ -19,7 +19,7 @@ open class AbstractPluginTest : AbstractIntegrationTest() {
             pluginManagement {
                 repositories {
                     maven { setUrl("$testRepository") }
```
bamboo: :o: does it still have to be `setUrl(...)` here?

eskatos: Oh, definitely no! Good catch!

PR #613, opened 2017-11-27
Remove usage of `DefaultIdeDependencyResolver`

bamboo:


tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinLibSources.kt
```diff
@@ -115,15 +115,5 @@ fun kotlinComponent(module: String): ComponentIdentifier =
...
-
-private
-fun moduleId(group: String, module: String, version: String) =
-    object : ModuleComponentIdentifier {
-        override fun getGroup() = group
-        override fun getModule() = module
-        override fun getVersion() = version
-        override fun getDisplayName() = "$group:$module:$version"
-    }
+fun moduleId(group: String, module: String, version: String): ModuleComponentIdentifier =
```
JLLeitschuh: Isn't `version` nullable in Gradle? Maybe this is a different thing and I'm conflating the two.

tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinLibSources.kt
```diff
@@ -58,20 +58,18 @@ fun sourcePathFor(scriptHandlers: List<ScriptHandler>): ClassPath {
...
-    resolvedDependencies.containsAll(
-        builtinKotlinModules.map(::kotlinModuleVersionIdentifier))
-
-
-private
-fun classpathDependenciesOf(buildscript: ScriptHandler): List<ModuleVersionIdentifier> =
-    DefaultIdeDependencyResolver()
-        .getIdeRepoFileDependencies(buildscript.configurations[CLASSPATH_CONFIGURATION])
-        .map { it.id }
+fun containsBuiltinKotlinModules(resolvedDependencies: HashSet<ComponentIdentifier>) =
```
JLLeitschuh: Why `HashSet` instead of just a `Set`? is there some specific behaviour of a `HashSet` you need??

tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinLibSources.kt
bamboo: It can't be null in this context by definition.

tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinLibSources.kt
bamboo: It's just an artifact of the refactoring.

tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinLibSources.kt
JLLeitschuh: What is the `ModuleComponentIdentifier`?
The [javadocs](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/component/ModuleComponentIdentifier.html) don't really do a good job explaining what it is.

tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinLibSources.kt
```diff
@@ -115,15 +116,5 @@ fun kotlinComponent(module: String): ComponentIdentifier =
...
-private
-fun moduleId(group: String, module: String, version: String) =
-    object : ModuleComponentIdentifier {
-        override fun getGroup() = group
-        override fun getModule() = module
-        override fun getVersion() = version
-        override fun getDisplayName() = "$group:$module:$version"
-    }
+fun moduleId(group: String, module: String, version: String): ModuleComponentIdentifier =
+    DefaultModuleComponentIdentifier.newId(group, module, version)
```
oehme: This is an internal class (and the interface is not meant for user extension either). Could you replace it with a predicate that checks the identifiers in `containsBuiltinKotlinModules`?

bamboo: That makes sense, @oehme, thanks.

PR #616, opened 2017-11-28
Add sample demonstrating Kapt plugin usage with Kotlin DSL

gildor:
### Context
Sample of usage [kapt](https://kotlinlang.org/docs/reference/kapt.html) with Kotlin DSL.
I think it can be useful because we already have Kotlin sample and Kapt doesn't provide friendly DSL for javac and kapt option configuration.
Example uses AutoValue for demonstration of code generation

### Contributor Checklist


JLLeitschuh:
This is really cool. I didn't know you could use auto-value like this. Great to see this sample just for general Gradle knowledge.

samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/HelloKaptSampleTest.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.junit.Assert.assertThat
+import org.junit.Test
+
+
+class HelloKaptSampleTest : AbstractSampleTest("hello-kapt") {
+
+    @Test
+    fun `hello kapt`() {
+        val output = build("run").output
+        println(output)
```
eskatos: ❌ Could you remove this `println()` statement?

samples/hello-kapt/build.gradle.kts
```diff
@@ -0,0 +1,54 @@
...
+        option("SomeJavacOption", "OptionValue")
+    })
+
+    //Kapt Arguments configuration
+    arguments(delegateClosureOf<KaptAnnotationProcessorOptions> {
+        arg("SomeKaptArgument", "ArgumentValue")
+    })
+}
+
+// Workaround for https://youtrack.jetbrains.com/issue/KT-17923 and related issues
```
eskatos: :x: While this may be valuable to people using "old" Kotlin versions I think we should not keep this around in the sample

samples/hello-kapt/build.gradle.kts
```diff
@@ -0,0 +1,54 @@
...
+import org.gradle.kotlin.dsl.delegateClosureOf
+import org.jetbrains.kotlin.gradle.plugin.KaptAnnotationProcessorOptions
+import org.jetbrains.kotlin.gradle.plugin.KaptJavacOptionsDelegate
+import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
+
+plugins {
+    application
+    idea
+    kotlin("jvm") version "1.1.61"
+    //Enable kapt plugin
```
eskatos: :o: Unnecessary comment

samples/hello-kapt/build.gradle.kts
gildor: Okay,of course. I prepared this sample in times of Kotlin 1.1.51

samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/HelloKaptSampleTest.kt
gildor: Oh, sorry, forgotten debug

gildor:
@eskatos Did requested changes, rebased changes

eskatos:
@gildor thanks for the changes

Having a sample demonstrating `kapt` usage makes a lot of sense.

But I'm a bit reluctant to merge this as is for two reasons:
- using `AutoValue` in Kotlin doesn't make much sense as data classes would be better in this case, using something more useful to Kotlin code would be nicer ;
- the sample has to rely on `delegateClosureOf()`, this should be fixed in the upstream plugin to make it kotlin friendly.

I'd say the first point should be done before merging while the second one can be done as a follow up.

gildor:
@eskatos I see. I agree that in most cases AutoValue is redundant in Kotlin project, but still valid in some situations: builder support, java compatibility (data class copy method usage from java is very inconvenient because of lack of default arguments), plugins support, some validation cases.
Also AutoValue simple enough to show that code generation works.
I can rewrite example to Dagger for example if you think it makes more sense in terms of this sample.
If you know some other popular java/kotlin library that can be used in domain of Java, I would happy to use it for example instead.

Kapt DSL not friendly to Kotlin-DSL, it's true, so maybe, for now, example even more valuable.
I've asked Yan Zhulanow from JB about this, he asked to create an Issue.
I will do that and maybe even contribute to kapt plugin to fix that and update example when fixed kapt plugin will be available.

gildor:
@eskatos Issue in Kotlin tracker for Kapt plugin https://youtrack.jetbrains.com/issue/KT-21596

bamboo:
> but still valid in some situations: builder support

Demonstrating the builder support would be nice as it's something you don't get with Kotlin data classes.

gildor:
@eskatos @bamboo I've updated AutoValue example with a builder example that shows how to use builders for validation and value normalization (this is impossible now with data classes).
Also, I added unit tests to sample logic and Readme.

I can also provide a sample with [auto-value-gson](https://github.com/rharter/auto-value-gson) extension, that generates nullable safe adapter for Gson with default value support. It's useful library, but I decided that sample will be too complicated without knowledge of Gson (or other Java json mapping libraries) pitfalls.

samples/hello-kapt/build.gradle.kts
```diff
@@ -0,0 +1,46 @@
+import org.gradle.api.internal.HasConvention
+import org.gradle.kotlin.dsl.delegateClosureOf
+import org.jetbrains.kotlin.gradle.plugin.KaptAnnotationProcessorOptions
+import org.jetbrains.kotlin.gradle.plugin.KaptJavacOptionsDelegate
+import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
+
+plugins {
+    application
+    idea
```
eskatos: :x: I would remove the `idea` plugin from the sample

samples/hello-kapt/build.gradle.kts
```diff
@@ -0,0 +1,46 @@
+import org.gradle.api.internal.HasConvention
```
eskatos: :x: unused import

samples/hello-kapt/build.gradle.kts
```diff
@@ -0,0 +1,46 @@
...
+kapt {
+    // Type safe accessors for Kapt DSL
+    correctErrorTypes = true
+
+    // Example of Javac Options configuration
+    javacOptions(delegateClosureOf<KaptJavacOptionsDelegate> {
+        option("SomeJavacOption", "OptionValue")
+    })
+
+    //Kapt Arguments configuration
```
eskatos: :x: this comment and others above miss a space after `//`

samples/hello-kapt/build.gradle.kts
gildor: Will remove

samples/hello-kapt/build.gradle.kts
gildor: Yeah. Also strange, that Idea doesn't show this (and some other) import as unused

samples/hello-kapt/build.gradle.kts
gildor: Fixed

gildor:
Fixed and rebased

gildor:
@eskatos Looks like Kapt extensions fixed and will be available in one of future Kotlin Plugin versions.
Should we wait and merge sample with those changes?

eskatos:
Hi @gildor,
Thanks for the heads up.
I'd say we can merge this sample before that upstream change, it has value already.

Could you rebase this branch on top of `develop` and make the appropriate changes?
(mostly Gradle wrapper and Kotlin version I think).

gildor:
@eskatos Great!
I rebased my PR with develop (update Kotlin and Wrapper)

eskatos:
And merged. Thank you @gildor!

PR #638, opened 2017-12-20
Convert `buildSrc/build.gradle` to Kotlin

bamboo:
It is about time!


JLLeitschuh:
https://www.myinstants.com/instant/that-was-easy/

Issue #684, opened 2018-01-19
Support Kotlin based init scripts specified in the command line


bamboo:
E.g., `./gradlew -I my.init.gradle.kts`

bamboo:
Closing as delivered.

PR #687, opened 2018-01-20
Allow `Gradle` as a script target for init scripts

bamboo:
With an `initscript` block instead of a `buildscript` block.

See #684

### DIFF
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinInitScript.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinInitScript.kt
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.kotlin.dsl
+
+import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.invocation.Gradle
+
+import org.gradle.kotlin.dsl.resolver.KotlinBuildScriptDependenciesResolver
+
+import kotlin.script.extensions.SamWithReceiverAnnotations
+import kotlin.script.templates.ScriptTemplateDefinition
+
+
+/**
+ * Base class for Kotlin init scripts.
+ */
+@ScriptTemplateDefinition(
+    resolver = KotlinBuildScriptDependenciesResolver::class,
+    scriptFilePattern = ".+\\.init\\.gradle\\.kts")
+@SamWithReceiverAnnotations("org.gradle.api.HasImplicitReceiver")
+abstract class KotlinInitScript(gradle: Gradle) : Gradle by gradle {
+
+    /**
+     * Configures the classpath of the init script.
+     */
+    @Suppress("unused")
+    open fun initscript(@Suppress("unused_parameter") block: ScriptHandlerScope.() -> Unit) = Unit
+
+    /**
+     * Creates a [ConfigurableFileCollection] containing the given files.
+     *
+     * You can pass any of the following types to this method:
+     *
+     * - A [CharSequence], including [String] as defined by [KotlinSettingsScript.file].
+     * - A [File] as defined by [KotlinSettingsScript.file].
+     * - A [java.nio.file.Path] as defined by [KotlinSettingsScript.file].
+     * - A [URI] or [java.net.URL] as defined by [KotlinSettingsScript.file].
+     * - A [org.gradle.api.file.Directory] or [org.gradle.api.file.RegularFile]
+     *   as defined by [KotlinSettingsScript.file].
+     * - A [Sequence], [Array] or [Iterable] that contains objects of any supported type.
+     *   The elements of the collection are recursively converted to files.
+     * - A [org.gradle.api.file.FileCollection].
+     *   The contents of the collection are included in the returned collection.
+     * - A [org.gradle.api.provider.Provider] of any supported type.
+     *   The provider's value is recursively converted to files. If the provider represents an output of a task,
+     *   that task is executed if the file collection is used as an input to another task.
+     * - A [java.util.concurrent.Callable] that returns any supported type.
+     *   The callable's return value is recursively converted to files.
+     *   A `null` return value is treated as an empty collection.
+     * - A [org.gradle.api.Task].
+     *   Converted to the task's output files.
+     *   The task is executed if the file collection is used as an input to another task.
+     * - A [org.gradle.api.tasks.TaskOutputs].
+     *   Converted to the output files the related task.
+     *   The task is executed if the file collection is used as an input to another task.
+     * - Anything else is treated as a failure.
+     *
+     * The returned file collection is lazy, so that the paths are evaluated only when the contents of the file
+     * collection are queried. The file collection is also live, so that it evaluates the above each time the contents
+     * of the collection is queried.
+     *
+     * The returned file collection maintains the iteration order of the supplied paths.
+     *
+     * The returned file collection maintains the details of the tasks that produce the files,
+     * so that these tasks are executed if this file collection is used as an input to some task.
+     *
+     * This method can also be used to create an empty collection, which can later be mutated to add elements.
+     *
+     * @param paths The paths to the files. May be empty.
+     * @return The file collection.
+     */
+    @Suppress("unused")
+    fun files(vararg paths: Any): ConfigurableFileCollection =
+        fileOperations.files(paths)
+
+    private
+    val fileOperations by lazy {
+        fileOperationsFor(gradle, gradle.startParameter.currentDir)
+    }
+}
+
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinBuildScriptCompiler.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinBuildScriptCompiler.kt
@@ -16,6 +16,8 @@
 
 package org.gradle.kotlin.dsl.provider
 
+import org.gradle.api.initialization.dsl.ScriptHandler
+
 import org.gradle.api.internal.initialization.ClassLoaderScope
 import org.gradle.api.internal.initialization.ScriptHandlerInternal
 import org.gradle.api.internal.plugins.PluginAwareInternal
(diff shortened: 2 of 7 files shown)


provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinInitScript.kt
```diff
@@ -0,0 +1,95 @@
...
+     * @param paths The paths to the files. May be empty.
+     * @return The file collection.
+     */
+    @Suppress("unused")
+    fun files(vararg paths: Any): ConfigurableFileCollection =
+        fileOperations.files(paths)
+
+    private
+    val fileOperations by lazy {
+        fileOperationsFor(gradle, gradle.startParameter.currentDir)
```
JLLeitschuh: Is this the current directory of the `init.gradle.kts` file or the directory where gradle is invoked from?

provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinInitScript.kt
```diff
@@ -0,0 +1,95 @@
...
+
+
+/**
+ * Base class for Kotlin init scripts.
+ */
+@ScriptTemplateDefinition(
+    resolver = KotlinBuildScriptDependenciesResolver::class,
+    scriptFilePattern = ".+\\.init\\.gradle\\.kts")
+@SamWithReceiverAnnotations("org.gradle.api.HasImplicitReceiver")
+abstract class KotlinInitScript(gradle: Gradle) : Gradle by gradle {
```
eskatos: :x: Missing `ScriptApi` implementation and corresponding assertions in `ScriptApiTest`

provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinInitScript.kt
eskatos: :x: To be on par with the Groovy DSL implementation it should be:
- the directory of the init script if it is a regular file
- the base dir of the default `FileResolver` if the init script is not a regular file (e.g. remote URI)

See https://github.com/gradle/gradle/blob/010abecca658c2d9c92c2d223b8a7554eff0ccf4/subprojects/core/src/main/java/org/gradle/groovy/scripts/DefaultScript.java#L88-L93

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/SettingsExtensions.kt
```diff
@@ -17,7 +17,14 @@ package org.gradle.kotlin.dsl.support
...
+import org.gradle.api.invocation.Gradle
+
 
 inline
 fun <reified T : Any> Settings.serviceOf(): T =
+    gradle.serviceOf()
+
+
+inline
+fun <reified T : Any> Gradle.serviceOf(): T =
```
eskatos: :o: Should go to some `GradleExtensions.kt` file

provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinInitScript.kt
JLLeitschuh: So what you're saying is that this is incorrect as it is?

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/KotlinInitScriptIntegrationTest.kt
```diff
@@ -0,0 +1,30 @@
...
+                }
+
+                val computer = ${DeepThought::class.qualifiedName}()
+                val answer = computer.compute()
+                println("*" + answer + "*")
+            """)
+
+        assert(
+            build("-I", initScript.canonicalPath).output.contains("*42*"))
+    }
```
JLLeitschuh: Can you add a test that covers making sure that the `baseDir: File` is set in accordance to the behaviour of groovy based init scripts?

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptTarget.kt
```diff
@@ -76,6 +80,16 @@ fun projectScriptTarget(project: Project, topLevelScript: Boolean): KotlinScript
         })
 
 
+private
+fun gradleScriptTarget(gradle: Gradle) =
+    KotlinScriptTarget(
+        gradle,
+        type = Gradle::class,
```
JLLeitschuh: :o: Is this `Gradle` type used for anything other than Init scripts? If it is then this method might make sense to be named `gradleInitScriptTarget`. If not, this is probably fine.

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinBuildScriptCompiler.kt
```diff
@@ -232,6 +229,22 @@ class KotlinBuildScriptCompiler(
...
+        compiledScript: CachingKotlinCompiler.CompiledScript,
+        classLoaderScope: ClassLoaderScope,
+        action: (Class<*>) -> Unit) {
+
+        val scriptClass = classFrom(compiledScript, classLoaderScope)
+        withContextClassLoader(scriptClass.classLoader) {
+            try {
+                action(scriptClass)
+            } catch (e: InvocationTargetException) {
+                throw e.targetException
```
JLLeitschuh: Why are you dropping the outer `InvocationTargetException`?

bamboo: Because it adds no value.

JLLeitschuh: Okay. Fair enough. Just as long as no useful information is lost.

bamboo:
@eskatos This is ready for your review now.

provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinInitScript.kt
```diff
@@ -0,0 +1,384 @@
...
+    gradle: Gradle) : Gradle by gradle {
+
+    /**
+     * Configures the classpath of the init script.
+     */
+    @Suppress("unused")
+    open fun initscript(@Suppress("unused_parameter") block: ScriptHandlerScope.() -> Unit) = Unit
+
+    /**
+     * Logger for settings. You can use this in your settings file to write log messages.
```
eskatos: :x: This should read *init scripts* instead of *settings*

provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinInitScript.kt
```diff
@@ -0,0 +1,384 @@
...
+    /**
+     * Provides access to resource-specific utility methods, for example factory methods that create various resources.
+     */
+    @Suppress("unused")
+    val resources: ResourceHandler by lazy { fileOperations.resources }
+
+    /**
+     * Returns the relative path from this script's target base directory to the given path.
+     *
+     * The given path object is (logically) resolved as described for [KotlinSettingsScript.file],
```
eskatos: :x: `s/KotlinSettingsScript/KotlinInitScript/` and lots of other places below

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinScriptHost.kt
```diff
@@ -0,0 +1,57 @@
...
+
+import org.gradle.groovy.scripts.ScriptSource
+
+import org.gradle.internal.service.ServiceRegistry
+
+import org.gradle.kotlin.dsl.fileOperationsFor
+import org.gradle.kotlin.dsl.invoke
+
+
+class KotlinScriptHost(
```
eskatos: :x: This is now only used for init scripts and makes implementation for various targets different. Could this be used for other targets already? Doing this in a subsequent PR would be fine by me, but I'd like the implementations to be as similar as possible.

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/GradleExtensions.kt
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright 2017 the original author or authors.
```
JLLeitschuh: Slightly off topic: This year should probably be updated sometime soon.

provider/src/main/kotlin/org/gradle/kotlin/dsl/ActionExtensions.kt
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+
+package org.gradle.kotlin.dsl
+
+import org.gradle.api.Action
+
+
+/**
+ * Enables function invocation syntax on [Action] references.
```
JLLeitschuh: The function's existence enables this. But calling the function does not enable the ability to use the invocation syntax on action.
Consider changing the documentation to describe what the method is doing, not why this function exists.

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/KotlinInitScriptIntegrationTest.kt
```diff
@@ -0,0 +1,55 @@
...
+                println("*" + answer + "*")
+            """)
+
+        assert(
+            build("-I", initScript.canonicalPath)
+                .output.contains("*42*"))
+    }
+
+    @Test
+    fun `initscript file path is resolved against parent script dir`() {
```
JLLeitschuh: 🎉

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinScriptHost.kt
bamboo: Of course. But that's not related to supporting init scripts and it would change behaviour on the other targets so it is best handled as a separate issue.

provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinInitScript.kt
bamboo: Right. Thanks.

provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinInitScript.kt
bamboo: Right.

provider/src/main/kotlin/org/gradle/kotlin/dsl/ActionExtensions.kt
bamboo: Thanks for the suggestion, @JLLeitschuh. I like how it is. Feel free to open a PR if you feel strongly about it.

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinScriptHost.kt
eskatos: 👍 Sounds good

bamboo:
@eskatos, ready for another round.

Issue #683, opened 2018-01-19
Kotlin script compilation error reporting should be easy to read and actionable


eskatos:
The current script compilation error scenario is very confusing.

### Current Behavior

Given:
```kotlin
// build.gradle.kts
unresolvableReference
```

`./gradlew help` output is:
```
> Configure project :
e: /path/to/build/root/dir/build.gradle.kts:1:1: Unresolved reference: unresolvableReference


FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'kdsl-error-reporting'.
> Could not open cache directory absp8o35bf4821ved7vw9d420 (/path/to/gradle-user-home/.gradle/caches/4.6-20180118075735+0000/gradle-kotlin-dsl/absp8o35bf4821ved7vw9d420).
   > Internal error: unable to compile script, see log for details

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 1s
```

The reported error is `Could not open cache directory`, not useful.
The cause is `Internal error: unable to compile script, see log for details`, why "internal error"?
I need to go read the output above to find out what and where the error was.
If the compiler outputs a lot of messages, the one I'm interested in is probably the first error, finding it may require some effort.

Looking at build scans the [summary](https://scans.gradle.com/s/a3w6wrftnk4yw#failure) don't show any actionable data, nor do [failure details](https://scans.gradle.com/s/a3w6wrftnk4yw/failure?openFailures=WzBd&openStackTraces=WzIse31d#top=0). I have to look at the [console log](https://scans.gradle.com/s/a3w6wrftnk4yw/console-log#L1) to find the actionable data.

### Expected Behavior

With the same example the output should look like something along the lines of:

```
FAILURE: Build failed with an exception.

* Where:
Build file '/path/to/build/root/dir/build.gradle.kts' line: 1

* What went wrong:
Could not compile build file '/path/to/build/root/dir/build.gradle.kts'.
> Script body compilation error:
  e: /path/to/build/root/dir/build.gradle.kts:1:1: Unresolved reference: unresolvableReference
      unresolvableReference
      ^

  1 error

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 1s
```

[...]

PR #688, opened 2018-01-20
Make script compilation error output nicer

bamboo:
By showing the very first script compilation error prominently.

See #683

### DIFF
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinBuildScriptCompiler.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinBuildScriptCompiler.kt
@@ -24,9 +24,11 @@ import org.gradle.groovy.scripts.ScriptSource
 
 import org.gradle.internal.classpath.ClassPath
 import org.gradle.internal.classpath.DefaultClassPath
+import org.gradle.internal.exceptions.LocationAwareException
 
 import org.gradle.kotlin.dsl.get
 import org.gradle.kotlin.dsl.support.EmbeddedKotlinProvider
+import org.gradle.kotlin.dsl.support.ScriptCompilationException
 import org.gradle.kotlin.dsl.support.compilerMessageFor
 
 import org.gradle.plugin.management.internal.PluginRequests
@@ -210,27 +212,41 @@ class KotlinBuildScriptCompiler(
 
     private
     fun compileBuildscriptBlock(buildscriptRange: IntRange, scriptTemplate: KClass<*>) =
-        kotlinCompiler.compileBuildscriptBlockOf(
-            scriptTemplate,
-            scriptPath,
-            script.linePreservingSubstring(buildscriptRange),
-            buildscriptBlockCompilationClassPath)
+        withKotlinCompiler {
+            compileBuildscriptBlockOf(
+                scriptTemplate,
+                scriptPath,
+                script.linePreservingSubstring(buildscriptRange),
+                buildscriptBlockCompilationClassPath)
+        }
 
     private
     fun compilePluginsBlock(pluginsRange: IntRange, scriptTemplate: KClass<*>) =
-        kotlinCompiler.compilePluginsBlockOf(
-            scriptTemplate,
-            scriptPath,
-            script.linePreservingSubstring_(pluginsRange),
-            pluginsBlockCompilationClassPath)
+        withKotlinCompiler {
+            compilePluginsBlockOf(
+                scriptTemplate,
+                scriptPath,
+                script.linePreservingSubstring_(pluginsRange),
+                pluginsBlockCompilationClassPath)
+        }
 
     private
     fun compileScriptFileFor(classPath: ClassPath) =
-        kotlinCompiler.compileGradleScript(
-            scriptTarget.scriptTemplate,
-            scriptPath,
-            script,
-            classPath)
+        withKotlinCompiler {
+            compileGradleScript(
+                scriptTarget.scriptTemplate,
+                scriptPath,
+                script,
+                classPath)
+        }
+
+    private
+    fun <T> withKotlinCompiler(action: CachingKotlinCompiler.() -> T): T =
+        try {
+            kotlinCompiler.action()
+        } catch (e: ScriptCompilationException) {
+            throw LocationAwareException(e, scriptSource, e.errors.first().lineNumber)
+        }
 
     private
     fun classFrom(compiledScript: CachingKotlinCompiler.CompiledScript, scope: ClassLoaderScope): Class<*> =
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinBuildScriptCompiler.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinBuildScriptCompiler.kt
@@ -245,7 +245,7 @@ class KotlinBuildScriptCompiler(
         try {
             kotlinCompiler.action()
         } catch (e: ScriptCompilationException) {
-            throw LocationAwareException(e, scriptSource, e.errors.first().lineNumber)
+            throw LocationAwareException(e, scriptSource, e.firstErrorLine)
         }
 
     private
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
@@ -216,23 +216,21 @@ data class ScriptCompilationError(val message: String, val location: CompilerMes
 internal
 data class ScriptCompilationException(val errors: List<ScriptCompilationError>) : RuntimeException() {
     override val message: String?
-        get() =
-            if (errors.isEmpty()) "Script compilation failed, see the compiler logs above for more insights."
-            else {
-                val errorPlural = if (errors.size > 1) "errors" else "error"
-                val maxLineNumberLen = errors.mapNotNull { it.location?.line }.max().toString().length
-                (listOf("Script compilation $errorPlural:") + errors.map { error ->
-                    if (error.location != null) {
-                        "${error.location.line.toString().padStart(maxLineNumberLen, '0')}: ${error.location.lineContent}\n" +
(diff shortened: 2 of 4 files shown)


provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinBuildScriptCompiler.kt
```diff
@@ -210,27 +212,41 @@ class KotlinBuildScriptCompiler(
...
+                script,
+                classPath)
+        }
+
+    private
+    fun <T> withKotlinCompiler(action: CachingKotlinCompiler.() -> T): T =
+        try {
+            kotlinCompiler.action()
+        } catch (e: ScriptCompilationException) {
+            throw LocationAwareException(e, scriptSource, e.errors.first().lineNumber)
```
JLLeitschuh: Is a `ScriptCompilationException` guaranteed to always have at least one error?

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
```diff
@@ -204,44 +205,60 @@ fun kotlinCoreEnvironmentFor(configuration: CompilerConfiguration, rootDisposabl
...
+internal
+data class ScriptCompilationError(val message: String, val lineNumber: Int?)
 
-        override fun hasErrors() = errors > 0
 
-        override fun clear() { errors = 0 }
+internal
+data class ScriptCompilationException(val errors: List<ScriptCompilationError>) : RuntimeException() {
+    override val message: String?
+        get() = errors.first().message
```
JLLeitschuh: Same question as above:
https://github.com/gradle/kotlin-dsl/pull/688/files#r163057841

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
```diff
@@ -204,44 +205,60 @@ fun kotlinCoreEnvironmentFor(configuration: CompilerConfiguration, rootDisposabl
...
-                CompilerMessageSeverity.INFO -> log.info { msg() }
-                else -> log.debug { taggedMsg() }
+            } ?: message
+
+        fun taggedMsg() =
+            "${severity.presentableName[0]}: ${msg()}"
+
+        when (severity) {
+            CompilerMessageSeverity.ERROR, CompilerMessageSeverity.EXCEPTION -> {
+                errors += ScriptCompilationError(message, location?.line)
```
JLLeitschuh: Is `report` guaranteed to only be called from one thread or could it be called by two threads concurrently?
If it could be called by two threads concurrently you could get `ConcurrentModificationException`

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinBuildScriptCompiler.kt
eskatos: It is not guaranteed. Fixed

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
eskatos: It is not guaranteed. Fixed

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
eskatos: Good question.

@ligee do you know if Kotlin compiler's `MessageCollector` could be invoked concurrently?

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinBuildScriptCompiler.kt
bamboo: It is.

bamboo: I explain, `compileBunchOfSources` returns `false` if `messageCollector.hasErrors()`.

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
```diff
@@ -204,44 +205,76 @@ fun kotlinCoreEnvironmentFor(configuration: CompilerConfiguration, rootDisposabl
...
+data class ScriptCompilationError(val message: String, val location: CompilerMessageLocation?)
 
-        var errors = 0
 
-        override fun hasErrors() = errors > 0
+internal
+data class ScriptCompilationException(val errors: List<ScriptCompilationError>) : RuntimeException() {
+    override val message: String?
+        get() =
+            if (errors.isEmpty()) "Script compilation failed, see the compiler logs above for more insights."
```
bamboo: :x: This exception is only thrown when there are errors.

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinBuildScriptCompiler.kt
eskatos: Right, thanks! I'll follow up

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
```diff
@@ -602,11 +602,44 @@ class GradleKotlinDslIntegrationTest : AbstractIntegrationTest() {
...
+                FAILURE: Build failed with an exception.
+
+                * Where:
+                Build file '${buildFile.canonicalPath}' line: 1
+
+                * What went wrong:
+                Script compilation errors:
+
+                  01: println(foo)
+                              ^- Unresolved reference: foo
```
bamboo: 👍 Very nice!

JLLeitschuh: This is absolutely awesome! Wondering if it's helpful to have it say `Line 01: println(foo)` instead of just `01: println(foo)`

eskatos: @JLLeitschuh, I was wondering the same. Something is missing to explain what these numbers are.

```
Line 01: println(foo)
                 ^- Unresolved reference: foo
```

or, shorter:
```
L01: println(foo)
             ^- Unresolved reference: foo
```

JLLeitschuh: Either one is better than not having the information at all. It's really confusing just seeing numbers without the label. For all the user knows they could be error codes.

Reading the test was initially confusing until I figured out that they were line numbers. You want the information to be as immediately apparent as possible.

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
```diff
@@ -204,44 +205,74 @@ fun kotlinCoreEnvironmentFor(configuration: CompilerConfiguration, rootDisposabl
...
+internal
+data class ScriptCompilationException(val errors: List<ScriptCompilationError>) : RuntimeException() {
+    override val message: String?
+        get() {
+            val errorPlural = if (errors.size > 1) "errors" else "error"
+            val maxLineNumberLen = errors.mapNotNull { it.location?.line }.max().toString().length
+            return (listOf("Script compilation $errorPlural:") + errors.map { error ->
+                if (error.location != null) {
+                    "${error.location.line.toString().padStart(maxLineNumberLen, '0')}: ${error.location.lineContent}\n" +
+                        "${" ".repeat(maxLineNumberLen + 1 + error.location.column)}^- ${error.message}"
```
JLLeitschuh: I wish the "up" part of this arrow character was more pronounced, it would work perfectly here
⬑
Other potential alternatives would be these:
 - ⇪
 - ▲
 - ⇧

This is the site I'm getting these examples from:
http://xahlee.info/comp/unicode_arrows.html

This works fine as is, but I wish that there was a character that would fit this use case better.

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
eskatos: I opted for the verbose and explicit option.
```
Line 01: println(foo)
                 ^- Unresolved reference: foo
```

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
ligee: Now, generally - not, but we cannot guarantee that it remains like this in the future.

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
```diff
@@ -272,6 +278,16 @@ class LoggingMessageCollector(
             else -> log.debug { taggedMsg() }
         }
     }
+
+    /**
+     * Safely record a script compilation error.
+     * There'are no guarantee that `MessageCollector`s are not invoked concurrently.
+     */
```
JLLeitschuh: 👍  Thanks for looking into this and resolving it.

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
JLLeitschuh: 👍  I think this was the right call.

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
JLLeitschuh: Bikeshed: I might have used a concurrent data structure like a `CopyOnWriteArrayList` instead, but this works fine as well.

PR #693, opened 2018-01-24
Move from jre7/8 to jdk7/8

StefMa:
### Context
This will fix #690 

I've simply replaced all jre7/8 changed to jdk7/8 in the files I found in this commit https://github.com/gradle/kotlin-dsl/commit/9dd7074deefe9d6f5c2985bd9d6c67d22d88ac7f

Hope that was all 👍 

### Contributor Checklist
- [x] Base the PR against the `develop` branch
- [x] Make sure that all commits are [signed off](https://git-scm.com/docs/git-commit#git-commit---signoff) to indicate that you agree to the terms of [Developer Certificate of Origin](https://developercertificate.org/).
- [x] ~Provide integration tests to verify changes from a user perspective~
- [x] ~Provide unit tests to verify logic~
- [ ] Ensure that tests pass locally: `./gradlew check --parallel`

PR #700, opened 2018-01-31
Move from kotlin-stdlib-jre8 to kotlin-stdlib-jdk8

eskatos:
Since Kotlin 1.2.

See http://kotlinlang.org/docs/reference/whatsnew12.html#kotlin-standard-library-artifacts-and-split-packages

This PR is a follow up to #693

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -28,8 +28,8 @@ allprojects {
     version = "0.16.0-SNAPSHOT"
 }
 
-val publishedPluginsVersion by extra { "0.15.4" }
-val futurePluginsVersion = "0.15.5"
+val publishedPluginsVersion by extra { "0.15.5" }
+val futurePluginsVersion = "0.15.6"
 project(":plugins") {
     group = "org.gradle.kotlin"
     version = futurePluginsVersion
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/EmbeddedKotlinProvider.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/EmbeddedKotlinProvider.kt
@@ -58,15 +58,15 @@ val embeddedModules: List<EmbeddedModule> by lazy {
     // TODO:pm could be generated at build time
     val annotations = EmbeddedModule("org.jetbrains", "annotations", "13.0")
     val stdlib = embeddedKotlin("stdlib", listOf(annotations))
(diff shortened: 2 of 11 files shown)


provider/src/main/kotlin/org/gradle/kotlin/dsl/support/EmbeddedKotlinProvider.kt
```diff
@@ -34,7 +34,7 @@ import java.util.*
 
 
 private
-const val embeddedRepositoryCacheKeyVersion = 1
+const val embeddedRepositoryCacheKeyVersion = 2
```
JLLeitschuh: This is a cache busting mechanism right?
I think I probably understand what this is, but can you add a quick comment above it explaining what it is and why it's necessary.

Also, why is the bump necessary?

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
```diff
@@ -410,13 +410,13 @@ class GradleKotlinDslIntegrationTest : AbstractIntegrationTest() {
     }
 
     @Test
-    fun `build script can use jre8 extensions`() {
+    fun `build script can use jdk8 extensions`() {
 
         assumeJavaLessThan9()
```
JLLeitschuh: Has this been fixed yet such that this assumption is no longer necessary?

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/EmbeddedKotlinProvider.kt
eskatos: See https://github.com/gradle/kotlin-dsl/pull/693#issuecomment-360105473

build.gradle.kts
```diff
@@ -28,8 +28,8 @@ allprojects {
     version = "0.16.0-SNAPSHOT"
 }
 
-val publishedPluginsVersion by extra { "0.15.4" }
-val futurePluginsVersion = "0.15.5"
+val publishedPluginsVersion by extra { "0.15.5" }
```
bamboo: That means you already published the new versions?

eskatos: Yes

PR #702, opened 2018-02-02
Forcefully delete temporary folder upon test end

bamboo:
And fail the test if the folder cannot be deleted and it's not annotated with `@LeaksFileHandles`.


test-fixtures/src/main/kotlin/org/gradle/kotlin/dsl/fixtures/ForcefullyDeletedTemporaryFolder.kt
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.rules.TemporaryFolder
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+
+import kotlin.annotation.AnnotationTarget.FUNCTION
+
+
+/**
+ * A [TemporaryFolder] JUnit rule that fails the test if the
+ * the temporary folder cannot be deleted when the test finishes and
```
eskatos: :o: `the` duplicated with previous line

eskatos:
👍

Issue #696, opened 2018-01-30
Introduce cross-build in-memory compiled Kotlin script cache


eskatos:
Dogfooding the Kotlin DSL on the `gradle/gradle` build made apparent the fact that the Kotlin DSL provider doesn't have a cross-build in-memory compiled script cache and thus is slower than the Groovy DSL provider.

The regression can be seen [here](https://builds.gradle.org/viewLog.html?buildId=10626274&buildTypeId=Gradle_Check_PerformanceTestCoordinator&tab=buildResultsDiv) and the flamegraphs [here](https://builds.gradle.org/viewLog.html?buildId=10632445&buildTypeId=Gradle_Util_Performance_AdHocPerformanceScenarioLinux&tab=artifacts#!-1o4yfzkjpc80h,3grkdhtd6wwj).

We can clearly see that calculating the on-disk cache key is pretty expensive (`ClasspathUtil.getClasspath()` & hashing). This calculation wouldn't be needed with an in-memory cache built the same way as the Groovy DSL provider one.

A cross-build in-memory compiled script cache should be introduced in the Kotlin DSL provider. According to the flame graphs it may make the Kotlin DSL provider faster than the Groovy one.

JLLeitschuh:
Just trying to understand, are you saying that the compiled Kotlin scripts don't get cached in memory in the daemon currently?

bamboo:
Compiled Kotlin scripts don't get cached in memory currently.

JLLeitschuh:
Okay, thanks for the clarification.

JLLeitschuh:
Okay, thanks for the clarification.

eskatos:
[Ad-hoc performance test build in `gradle/gradle`](https://builds.gradle.org/viewLog.html?buildId=10847127&buildTypeId=Gradle_Util_Performance_AdHocPerformanceScenarioLinux&tab=buildLog&_focus=2398#_state=84,2395,2397) shows a 3.4% speedup. Keep in mind that this test evaluates 12 Kotlin scripts and 111 Groovy scripts. The on-disk compilation cache disappeared from the [flame graphs](https://builds.gradle.org/viewLog.html?buildId=10847127&buildTypeId=Gradle_Util_Performance_AdHocPerformanceScenarioLinux&tab=artifacts#!-1o4yfzkjpc80h,3grkdhtd6wwj,-1o4yfzkjpc80h,3grkdhtd6wwj).

eskatos:
[...]

Issue #607, opened 2017-11-23
Evaluation Dependency on a child project causes exception org.gradle.api.internal.initialization.DefaultClassLoaderScope@31920b27 must be locked before it can be used to compute a classpath!


NikolayMetchev:
Attached is a project that cannot be built but I am not sure exactly why it is breaking. A dependency on a child project (e.g. :parent:b) breaks gradle kotlin dsl.

### Expected Behavior
project to build

### Current Behavior
output of gradlew clean command:
* What went wrong:
A problem occurred configuring project ':a'.
> A problem occurred configuring project ':parent:b'.
   > org.gradle.api.internal.initialization.DefaultClassLoaderScope@31920b27 must be locked before it can be used to compute a classpath!

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

* Get more help at https://help.gradle.org

### Steps to Reproduce (for bugs)
[dependencyEvaluationBug.zip](https://github.com/gradle/kotlin-dsl/files/1498897/dependencyEvaluationBug.zip)
Attached is a zip file that can be used to reproduce the issue

### Your Environment
Windows 10
Java 1.8 update 152

JLLeitschuh:
Let me just start by saying that this is a pretty weird project setup.


I don't think that you need the `evaluationDependsOn(":parent:b")` line.

I'm not certain if this is enabled by default but you may want to try making sure that `org.gradle.configureondemand=false` is set.

Because of: https://github.com/gradle/gradle/issues/1630

NikolayMetchev:
In the attached project org.gradle.configureondemand=false (the default) and the bug still reproduces. I am aware of that other bug and this is not that.
I do need the evaluationDependsOn(":parent:b") line because evaluation is done alphabetically and I need b to be evaluated before a.

JLLeitschuh:
If I remove the `evaluationDependsOn(":parent:b")` then the project compiles fine.
I recommend that you depend upon artifacts produced by `:parent:b` using the gradle `configuration` artifacts instead of depending upon the evaluation of the entirety of `:parent:b` before `a` can run.

Do you have this problem in a project written in `groovy` or only in a project written in the Gradle Kotlin DSL?

NikolayMetchev:
[...]

PR #707, opened 2018-02-05
 Introduce in-memory Kotlin scripts class loading cache

eskatos:
See #696

### DIFF
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/CachingKotlinCompiler.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/CachingKotlinCompiler.kt
@@ -44,6 +44,14 @@ import kotlin.script.experimental.dependencies.ScriptDependencies
 
 
 internal
+data class CompiledScript<out T : Any>(val location: File, val className: String, val metadata: T)
+
+
+internal
+data class PluginsBlockMetadata(val lineNumber: Int)
+
+
+internal
 class CachingKotlinCompiler(
     private val scriptCache: ScriptCache,
     private val implicitImports: ImplicitImports,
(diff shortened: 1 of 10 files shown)


provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/CachingKotlinCompiler.kt
```diff
@@ -43,6 +45,26 @@ import kotlin.script.experimental.dependencies.DependenciesResolver
 import kotlin.script.experimental.dependencies.ScriptDependencies
 
 
+internal
+data class ScriptBlock<out T : Any>(
```
JLLeitschuh: Is the `: Any` just to constrain this to a non null value?

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptClassloadingCache.kt
```diff
@@ -0,0 +1,115 @@
...
+    override fun equals(other: Any?): Boolean {
+        if (this === other) {
+            return true
+        }
+        if (other == null || this::class != other::class) {
+            return false
+        }
+        val key = other as ScriptCacheKey
+        return (parentClassLoader.get() != null && key.parentClassLoader.get() != null
+            && parentClassLoader.get() == key.parentClassLoader.get()
```
JLLeitschuh: The GC can't clear this reference between these two checks? If it can, you should store the loaders as local variables before comparing them.

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/ScriptCachingIntegrationTest.kt
```diff
@@ -0,0 +1,399 @@
...
+        @JvmStatic
+        val isolatedTestKitDir = IsolatedTestKitDir()
+    }
+
+    @Test
+    fun `same script, target type & classpath`() {
+
+        // given: multi-project build with same build files
+        val sameContent = """println("Same script content on ${'$'}this")"""
+        withMultiProjectBuild(left = sameContent, right = sameContent).apply {
```
bamboo: 👍 Nice DSL!

eskatos:
@bamboo, this is ready for another round

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptClassloadingCache.kt
```diff
@@ -0,0 +1,118 @@
...
+    override fun equals(other: Any?): Boolean {
+        if (this === other) {
+            return true
+        }
+        if (other == null || this::class != other::class) {
+            return false
+        }
+        val key = other as ScriptCacheKey
+        val thisParentLoader = parentClassLoader.get()
+        val otherParentLoader = key.parentClassLoader.get()
```
JLLeitschuh: 👍

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptClassloadingCache.kt
```diff
@@ -0,0 +1,118 @@
...
+    private
+    fun classFrom(compiledScript: CompiledScript<*>, scope: ClassLoaderScope): Class<*> =
+        classLoaderFor(compiledScript.location, scope)
+            .loadClass(compiledScript.className)
+
+    private
+    fun classLoaderFor(location: File, scope: ClassLoaderScope) =
+        scope
+            .local(DefaultClassPath(location))
+            .lock()
```
JLLeitschuh: What's with this call to `lock`? Does something need to be unlocked after calling it?

eskatos: Nope. See [ClassLoaderScope.lock()](https://github.com/gradle/gradle/blob/f001402c8b7f404c923d972f1bc524b2de7b3fcd/subprojects/core/src/main/java/org/gradle/api/internal/initialization/ClassLoaderScope.java#L94-L99)

JLLeitschuh: I know that this premature locking has caused issues before with the `org.gradle.configureondemand=true` set.
Has this been tested with that?

Related: #311

eskatos: This isn't new code, this PR simply moves it.
But yeah, this is related to #311 and #607

PR #724, opened 2018-02-21
Prefer plugin portal over repo.gradle.org

eskatos:
Because of https://youtrack.jetbrains.com/issue/KT-22974

### DIFF
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -9,7 +9,7 @@ buildscript {
     }
 
     configure(listOf(repositories, project.repositories)) {
-        maven(url = "https://repo.gradle.org/gradle/repo")
+        gradlePluginPortal()
     }
 }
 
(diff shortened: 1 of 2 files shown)


bamboo:
Is that better than fixing `repo.gradle.org`?

eskatos:
@bamboo Fixing `repo.gradle.org` would mean making it mirror the plugin portal, right?
This could be considered.
But making our build consume the plugin portal instead makes sense to me.

eskatos:
I also guess that "fixing `repo.gradle.org` won't fix our present issue because of the checksums mismatch.

bamboo:
By fixing `repo.gradle.org` I mean making it serve the right Kotlin artifacts, however we do it.

It doesn't feel right to build against the plugin portal for a few reasons but mainly because we can depend on dev versions with `repo.gradle.org` but not with the portal as the main repo.

PR #726, opened 2018-02-21
Do not require a sign-off from org members

eskatos:
This needs to be on the `master` branch for DCO to pick it up.

### DIFF
--- a/.github/dco.yml
+++ b/.github/dco.yml
@@ -1,0 +1,3 @@
+# Disable sign-off chcecking for members of the Gradle GitHub organization
+require:
+  members: false


.github/dco.yml
```diff
@@ -0,0 +1,3 @@
+# Disable sign-off chcecking for members of the Gradle GitHub organization
```
wolfs: You may want to correct the typo I introduced ;)

JLLeitschuh:
I'm not really certain what you're trying to get by having this DCO thing. But it seems like unnecessary hurdles for someone wanting to contribute.
What's the intended goal for using DCO?

@eriwen?

eskatos:
@JLLeitschuh see https://groups.google.com/forum/#!msg/gradle-dev/l3laB2vJRL8/5wuL5jh5AgAJ

eriwen:
@JLLeitschuh A primary goal here is to avoid the need for new contributors to sign a CLA. Do you find it too cumbersome?

JLLeitschuh:
My company already signed the CLA so I'm not sure why I also need to do this as well.
Perhaps there's a way to configure my git repository to auto-sign off when I commit so I won't have to deal with it in the future.

PR #732, opened 2018-02-23
Made Provider API sample more idiomatic for what we can do with Kotlin

### LINKED ISSUE #597: Provider API recommendations and differences from Groovy DSL
This was a discussion that came up WRT the Provider API documentation we're adding in 4.4: https://github.com/gradle/gradle/pull/3478

We're recommending that plugin authors expose a single getter for their `Property` (or `Provider`) properties. 

From the Groovy DSL, we generate methods to allow people to use `=` in build scripts with `Property`s:

```
class MyTask extends DefaultTask {
    @Input Property<String> someProperty = ...
}

task myTask(type: MyTask) {
   someProperty = "foo"
}
```

Under the covers, this is calling `someProperty.setFromAnyValue(...)`, which eventually calls `Property.set(T)` or `Property.set(Provider)`.

I know this isn't quite possible from the Kotlin DSL.  @eskatos pointed me to the PropertyStateExtensions and sample https://github.com/gradle/kotlin-dsl/blob/master/samples/provider-properties/build.gradle.kts

This is pretty close, but it has some differences/problems:
- Non-Kotlin plugins can't really expose a Kotlin property, so how people use a `Property` in the Kotlin DSL will have to be different
- Kotlin plugin authors need to write some boilerplate and come up with clever names (message vs messageProperty)
- Non-Kotlin plugins configuring Kotlin plugins don't have access to the `Property`.
- Groovy DSL don't have access to the `Property` (so they can't use the assignment syntax above)
- `@Optional` doesn't play nice with the PropertyStateExtensions.  

I don't think the idiomatic way for writing a Gradle plugin in Kotlin needs to be exactly the same as any other language, but we should consider the impact to the Groovy DSL, other plugins (not written in Kotlin) and the general confusion users may have if `Property`s are handled differently depending on which language the plugin was written in.

big-guy:
This takes the approach to make Kotlin look more like the explicitness
of Java until we can do something fancier (like setter overloads).

- We want to expose a single getter that returns the Provider/Property
- The GreetingExtension and Greeting task had unnecessarily mutable variables
- Handwritten separate setters/getters for a Property isn't what we want people
  to do.  And I think we don't want authors to expose special methods for
  setting Provider or raw values.

Fixes #597

### Contributor Checklist
- [ ] Base the PR against the `develop` branch
- [ ] Provide integration tests to verify changes from a user perspective
- [ ] Provide unit tests to verify logic
- [ ] Ensure that tests pass locally: `./gradlew check --parallel`

### DIFF
--- a/samples/provider-properties/build.gradle.kts
+++ b/samples/provider-properties/build.gradle.kts
@@ -1,8 +1,10 @@
 apply<GreetingPlugin>()
 
 configure<GreetingPluginExtension> {
-    message = "Hi from Gradle"
-    outputFiles = files("$buildDir/a.txt", "$buildDir/b.txt")
+    message.set("Hi from Gradle")
+    outputFiles.from(
+        project.layout.buildDirectory.file("a.txt"),
+        project.layout.buildDirectory.file("b.txt"))
 }
 
 open class GreetingPlugin : Plugin<Project> {
@@ -11,49 +13,38 @@ open class GreetingPlugin : Plugin<Project> {
 
         // Add the 'greeting' extension object
         val greeting = project.extensions.create(
-            "greeting",
-            GreetingPluginExtension::class.java,
-            project)
+                "greeting",
+                GreetingPluginExtension::class.java,
+                project)
 
         // Add a task that uses the configuration
         project.tasks {
             "hello"(Greeting::class) {
                 group = "Greeting"
-                provideMessage(greeting.messageProvider)
-                outputFiles = greeting.outputFiles
+                message.set(greeting.message)
+                outputFiles.setFrom(greeting.outputFiles)
             }
         }
     }
 }
 
 open class GreetingPluginExtension(project: Project) {
-
-    private
-    val messageProperty = project.objects.property<String>()
-
-    var message by messageProperty
-
-    val messageProvider: Provider<String> get() = messageProperty
-
-    var outputFiles by project.files()
+    val message = project.objects.property<String>()
+    val outputFiles: ConfigurableFileCollection = project.files()
 }
 
 open class Greeting : DefaultTask() {
-
-    private
-    val messageProperty = project.objects.property<String>()
-
     @get:Input
-    var message by messageProperty
+    val message = project.objects.property<String>()
 
     @get:OutputFiles
-    var outputFiles by project.files()
-
-    fun provideMessage(message: Provider<String>) = messageProperty.set(message)
+    val outputFiles: ConfigurableFileCollection = project.files()
 
     @TaskAction
     fun printMessage() {
-        logger.info("Writing message '$message' to files ${outputFiles.files}")
+        val message = message.get()
+        val outputFiles = outputFiles.files
+        logger.info("Writing message '$message' to files $outputFiles")
         outputFiles.forEach { it.writeText(message) }
     }
 }
--- a/samples/provider-properties/build.gradle.kts
+++ b/samples/provider-properties/build.gradle.kts
@@ -1,24 +1,28 @@
 apply<GreetingPlugin>()
 
+fun buildFile(path: String) = layout.buildDirectory.file(path)
+
 configure<GreetingPluginExtension> {
+
     message.set("Hi from Gradle")
+
     outputFiles.from(
-        project.layout.buildDirectory.file("a.txt"),
-        project.layout.buildDirectory.file("b.txt"))
+        buildFile("a.txt"),
+        buildFile("b.txt"))
 }
 
 open class GreetingPlugin : Plugin<Project> {
 
-    override fun apply(project: Project) {
+    override fun apply(project: Project): Unit = project.run {
 
         // Add the 'greeting' extension object
-        val greeting = project.extensions.create(
-                "greeting",
-                GreetingPluginExtension::class.java,
-                project)
+        val greeting = extensions.create(
+            "greeting",
+            GreetingPluginExtension::class.java,
+            project)
 
         // Add a task that uses the configuration
-        project.tasks {
+        tasks {
             "hello"(Greeting::class) {
                 group = "Greeting"
                 message.set(greeting.message)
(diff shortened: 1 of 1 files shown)


big-guy:
@eskatos WDYT of this? 

Short of Kotlin getting https://youtrack.jetbrains.com/issue/KT-4075, I think this is what our samples should look like to follow our recommendations for all of the other languages.  We don't want people writing boilerplate or creating separate names for values and providers because we may be able to generate something better later, it makes using a Property in Kotlin complicated (multiple names, different API) and it causes interoperability weirdness across languages.  Someone configuring a Property shouldn't care if the underlying task is written in Kotlin, Groovy or Java.  

This probably means we should also get rid of the `ProjectStateExtensions` since those extensions are only necessary for the `var something by someProvider` syntax.

JLLeitschuh:
This is annoying, but I understand why it's necessary.

samples/provider-properties/build.gradle.kts
```diff
@@ -1,8 +1,10 @@
 apply<GreetingPlugin>()
 
 configure<GreetingPluginExtension> {
-    message = "Hi from Gradle"
-    outputFiles = files("$buildDir/a.txt", "$buildDir/b.txt")
+    message.set("Hi from Gradle")
+    outputFiles.from(
```
bamboo: :x: I really don't see how this is an improvement over the original.

big-guy: Ah, I may have misunderstood what this was doing before.  It looked like there were two non-final `outputFiles` fields where we really want a single instance for the extension and another for the task.  But it seems like there is some kind of magic happening under the covers.

What does this look like from Java/Groovy?  If we leave it as it was and in Java I do...

```
extension.getOutputFiles().from("somewhere")
hello.getOutputFiles().from("anotherplace")
```

Does `extension.getOutputFiles()` have both "somewhere" and "anotherplace"?

bamboo: Right, there is some magic provided by the delegated property which transforms the assignment into a `setFrom` invocation, see this test case https://github.com/gradle/kotlin-dsl/blob/cb44112374e36b41732ab390531b8bc29e8de327/provider/src/test/kotlin/org/gradle/kotlin/dsl/ConfigurableFileCollectionExtensionsTest.kt#L17

Which works thanks to https://github.com/gradle/kotlin-dsl/blob/cb44112374e36b41732ab390531b8bc29e8de327/provider/src/main/kotlin/org/gradle/kotlin/dsl/ConfigurableFileCollectionExtensions.kt#L38-L37

Thus the original task configuration code:
```kotlin
outputFiles = greeting.outputFiles
```

It's semantically equivalent to the new:
```Kotlin
outputFiles.setFrom(greeting.outputFiles)
```

It trades some verbosity in the API provider for a terser DSL.

JLLeitschuh:
@bamboo I think that something that would be really useful would be adding some really thorough documentation regarding what each of these fields in this example is for and how they are supposed to be used.

I think I understand what's going on here but as evidenced by this PR and the ensuing conversation, there's some serious confusion.

> Handwritten separate setters/getters for a Property isn't what we want people
to do.
I agree this is something to avoid but given the constraints of the language, I don't think this can be avoided.

> And I think we don't want authors to expose special methods for setting Provider or raw values.
@big-guy I think that's exactly what you've done in this PR.

If this is the best practice pattern that plugin developers are expected to stamp out, there should be some serious explanations/examples behind how to craft your extension classes to provide the API's desired.

samples/provider-properties/build.gradle.kts
```diff
@@ -1,8 +1,10 @@
...
+        project.layout.buildDirectory.file("a.txt"),
```
bamboo: :x: `project` can be omitted.

big-guy:
@JLLeitschuh, just to circle back to this to address your comments:

> I think I understand what's going on here but as evidenced by this PR and the ensuing conversation, there's some serious confusion.

To be fair, when this sample was made, we were writing those sort of getters/setters in gradle/gradle too. We settled on the single getter way later and needed to come back to update this sample based on the experience of doing it the other way.

>> And I think we don't want authors to expose special methods for setting Provider or raw values.

> @big-guy I think that's exactly what you've done in this PR.

Sorry, I wasn't clear. 

This PR removed all of the handwritten methods to get or set the value of the `Property` (delegated `message` Kotlin property, `messageProvider`, `provideMessage`).  By "special methods for setting Provider", I meant those kinds of methods.  `provideMessage` was a handwritten setter to call `Property.set(Provider)` and `messageProvider` was a handwritten getter to get a `Provider` from the private `messageProperty`.  Like wise, `message` was a special setter for calling `Property.set(String)` and a special getter for `Property.get()`.

We want tasks and extensions to expose a single `Property` or `Provider` method (a getter) for a given property. All `get()`s/`set()`s go through that method. Any extra DSL magic we can mix-in and evolve later.

> If this is the best practice pattern that plugin developers are expected to stamp out, there should be some serious explanations/examples behind how to craft your extension classes to provide the API's desired.

We show the use of a `Property` in [this Gradle Plugin development guide](https://guides.gradle.org/implementing-gradle-plugins/#extensions_vs_conventions) and lay out [guidelines](https://docs.gradle.org/current/userguide/lazy_configuration.html#sec:lazy_configuration_faqs) in the User Manual.  We could always do better though.

JLLeitschuh:
So the implications of this decision are that as plugin API developers, we will not be able to expose an API like this to Kotlin DSL users:
```
configure<GreetingPluginExtension> {
    message = "Hi from Gradle"
}
```
Instead, if using the Kotlin DSL API consumers will have to do:
```
configure<GreetingPluginExtension> {
    message.set("Hi from Gradle")
}
```

Doesn't that go against the Groovy API Parity story?

It seems like the implications of this are that writing a Groovy API in Kotin will force authors to provide a less than optimal API experience to consumers.

bamboo:
@JLLeitschuh Removing the delegated property from the declared API of both the extension and the task does make the Kotlin configuration code feel less idiomatic. This is not the end of the story. though We're taking a step backwards so we can contemplate the big picture and decide how to improve the situation across the board. Stay tuned.

Issue #736, opened 2018-02-27
buildscript.sourceFile in applied scripts should return the applied script file


eskatos:
it currently returns the applying script file.

```kotlin
// other.gradle.kts
println("other: ${buildscript.sourceFile}")

// build.gradle.kts
println("main: ${buildscript.sourceFile}")
apply { from("other.gradle.kts") }
```

outputs:
```
main: /path/to/build.gradle.kts
other: /path/to/build.gradle.kts 
```

it should output this instead:
```
main: /path/to/build.gradle.kts
other: /path/to/other.gradle.kts 
```

JLLeitschuh:
Is this an issue specific to the Kotlin DSL or all applied scripts?

eskatos:
@JLLeitschuh it is specific to Kotlin DSL scripts.

eskatos:
It happens because `buildscript` returns the script target `ScriptHandler`.
It should return the script `ScriptHandler` instead.

bamboo:
Closing as delivered.

PR #737, opened 2018-02-28
Let scripts ScriptHandler be the one belonging to the script

eskatos:
See #736

### DIFF
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinBuildScript.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinBuildScript.kt
@@ -18,6 +18,7 @@ package org.gradle.kotlin.dsl
 
 import org.gradle.api.Project
 
+import org.gradle.api.initialization.dsl.ScriptHandler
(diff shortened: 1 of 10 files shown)


provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinBuildScript.kt
```diff
@@ -34,7 +35,13 @@ import kotlin.script.templates.ScriptTemplateDefinition
     scriptFilePattern = ".*\\.gradle\\.kts")
 @SamWithReceiverAnnotations("org.gradle.api.HasImplicitReceiver")
 @GradleDsl
-abstract class KotlinBuildScript(project: Project) : Project by project {
+abstract class KotlinBuildScript(
+    project: Project,
+    private val scriptHandler: ScriptHandler) : Project by project {
```
bamboo: :x: I would have added it to the `KotlinScriptHost` and passed the `host` to all script templates instead. This would pave the way for uniform `ScriptApi` behaviour without the need for additional arguments.

Issue #637, opened 2017-12-19
Applying external build script fails if url does not end with .gradle.kts


tomreay:
When applying an external build script via URL writen in Kotlin the script is applied as a Groovy script rather than a Kotlin script when there are query parameters in the URL.

### Expected Behavior
The script type is determined and the file is compiled as a Kotlin file

### Current Behavior
The script is compiled as a Groovy file

### Context
This occurs when the URL does not end with the extension, ie there are query parameters.  The specific case here is pre-signed S3 URLs.
Working: `https://files.example.com/external.gradle.kts`
Not working: `https://files.example.com/external.gradle.kts?someQueryParam=true`

### Steps to Reproduce (for bugs)
1. Create a kotlin build script
2. Upload the file to a remote server
3. Apply the file using `apply { from("https://files.example.com/external.gradle.kts?someQueryParam=true") }`
4. File is compiled as Groovy

bamboo:
Thanks for the report!

eskatos:
I confirmed the bug up to Gradle 4.5-rc-1.

This happens because `ScriptSource#getFileName()` returns the whole URL string, including query string and fragment, thus preventing proper script factory selection in `ScriptPluginFactorySelector`.

eskatos:
After further investigation, there's another thing to consider. The file name used by the file backed cache of Kotlin scripts compilation comes from the same function. And the Kotlin compiler chokes on the file name:

```
e: Source entry is not a Kotlin file: /path/to/cache/script.gradle.kts?query=string
```

PR #744, opened 2018-03-06
Compile Kotlin DSL scripts from regular Kotlin source sets

bamboo:


bamboo:
@eskatos ready for review.

provider/src/main/kotlin/org/gradle/kotlin/dsl/precompile/PrecompiledProjectScript.kt
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Script template definition for precompiled Kotlin script with a [Project] target.
+ */
+@ScriptTemplateDefinition(
+    resolver = PrecompiledScriptDependenciesResolver::class,
+    scriptFilePattern = "^.*\\.gradle\\.kts$")
+@SamWithReceiverAnnotations("org.gradle.api.HasImplicitReceiver")
+@GradleDsl
+abstract class PrecompiledProjectScript(project: Project) : Project by project {
```
JLLeitschuh: A comment about why this is needed and why it's different than the `ProjectScript` would be really useful.
Eg. 
```
These scripts are applied to project scripts.
```

Also why are they called "Precompiled"?

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/KotlinSettingsScriptIntegrationTest.kt
```diff
@@ -0,0 +1,34 @@
...
+        withSettings("""
+            apply { from("gradle/my.settings.gradle.kts") }
+        """)
+
+        withBuildScript("""
+            val answer: String by extra
+            println("*" + answer + "*")
+        """)
+
+        assert(build().output.contains("*42*"))
```
JLLeitschuh: :o Are you sure you want to use `assert` instead of something like Hamkrest's `assertThat("blah", contains("*42*"))`
[Link.](https://github.com/npryce/hamkrest/blob/9b66019c4c84e1aee1f2966b5f47f2d5f909f8d1/src/main/kotlin/com/natpryce/hamkrest/string_matchers.kt#L108)

test-fixtures/src/main/kotlin/org/gradle/kotlin/dsl/fixtures/Testing.kt
```diff
@@ -19,14 +24,20 @@ fun <T : Throwable> assertFailsWith(exception: KClass<out T>, block: () -> Unit)
...
 inline
 fun <reified T> assertInstanceOf(o: Any): T {
     assertThat(o, instanceOf(T::class.java))
     return o as T
 }
 
+
+fun classLoaderFor(vararg classPath: File): URLClassLoader =
+    URLClassLoader.newInstance(
+        classPath.map { it.toURI().toURL() }.toTypedArray())
```
JLLeitschuh: `URLClassLoader` has some new behaviour under JDK 9. Has this been tested with JDK 9?

provider/src/main/kotlin/org/gradle/kotlin/dsl/precompile/PrecompiledProjectScript.kt
bamboo: They are called "Precompiled" because those script templates are for Kotlin scripts compiled ahead of time as part of regular Kotlin sourcesets and then distributed as regular binary plugins. Does the name make more sense now?

test-fixtures/src/main/kotlin/org/gradle/kotlin/dsl/fixtures/Testing.kt
bamboo: The whole test-suite is continuously tested against JDK 9.

provider/src/main/kotlin/org/gradle/kotlin/dsl/precompile/PrecompiledProjectScript.kt
```diff
@@ -0,0 +1,69 @@
...
+import kotlin.script.extensions.SamWithReceiverAnnotations
+import kotlin.script.templates.ScriptTemplateDefinition
+
+
+/**
+ * Script template definition for precompiled Kotlin script with a [Project] target.
+ */
+@ScriptTemplateDefinition(
+    resolver = PrecompiledScriptDependenciesResolver::class,
+    scriptFilePattern = "^.*\\.gradle\\.kts$")
```
JLLeitschuh: What about URL script with query parameters? #637
Or is this the wrong place to fix this?

bamboo: Wrong place to fix it and this template would never be applied to a remote script.

provider/src/main/kotlin/org/gradle/kotlin/dsl/precompile/PrecompiledProjectScript.kt
JLLeitschuh: Yes, documenting that would be useful.

test-fixtures/src/main/kotlin/org/gradle/kotlin/dsl/fixtures/Testing.kt
JLLeitschuh: 👍

provider/src/main/kotlin/org/gradle/kotlin/dsl/precompile/PrecompiledProjectScript.kt
JLLeitschuh: Okay, I was confusing these with the applied scripts.

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/KotlinSettingsScriptIntegrationTest.kt
bamboo: 👍

Issue #668, opened 2018-01-17
Expose implicit imports to `*.gradle.kts` from regular Kotlin source sets


bamboo:

bamboo:
Closing as delivered in 54b0e5afebe078f21f34b8c7c6efec54168fc300.

PR #752, opened 2018-03-14
Expose implicit imports to precompiled scripts

bamboo:
See #668

### DIFF
--- a/plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslPlugin.kt
+++ b/plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslPlugin.kt
@@ -18,14 +18,8 @@ package org.gradle.kotlin.dsl.plugins.dsl
 import org.gradle.api.Plugin
 import org.gradle.api.Project
 
-import org.gradle.kotlin.dsl.*
+import org.gradle.kotlin.dsl.gradleKotlinDsl
 import org.gradle.kotlin.dsl.plugins.embedded.EmbeddedKotlinPlugin
-
-import org.gradle.kotlin.dsl.precompile.PrecompiledInitScript
-import org.gradle.kotlin.dsl.precompile.PrecompiledProjectScript
-import org.gradle.kotlin.dsl.precompile.PrecompiledSettingsScript
-
-import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 import org.jetbrains.kotlin.samWithReceiver.gradle.SamWithReceiverExtension
 import org.jetbrains.kotlin.samWithReceiver.gradle.SamWithReceiverGradleSubplugin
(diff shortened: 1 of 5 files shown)


plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslPlugin.kt
```diff
@@ -73,16 +84,57 @@ open class KotlinDslPlugin : Plugin<Project> {
...
     private
-    fun Project.configureCompilerTasks() {
+    fun Project.enablePrecompiledScripts() {
         tasks {
-            "compileKotlin"(KotlinCompile::class) {
-                kotlinOptions {
-                    freeCompilerArgs += listOf("-script-templates", scriptTemplates)
+
+            val implicitImportsFile = kotlinDslBuildDir().map { it.file("implicit-imports") }
+            val generateImplicitImportsFile by creating {
```
eskatos: ❓ this task will be executed for each source set that contains precompiled scripts. But the content of this file is bound to `(Gradle.version + KotlinDsl.version)`. Couldn't we generate this import file in some cache with a wider scope instead?

plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslPlugin.kt
```diff
@@ -73,16 +84,57 @@ open class KotlinDslPlugin : Plugin<Project> {
...
+                inputs.property("gradleVersion", gradleVersion())
+                inputs.property("gradleKotlinDslVersion", gradleKotlinDslVersion())
+                outputs.file(implicitImportsFile)
```
eskatos: :x: if keeping this task it should be made cacheable

plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslPlugin.kt
```diff
@@ -73,16 +84,57 @@ open class KotlinDslPlugin : Plugin<Project> {
...
             }
+
+            afterEvaluate {
+                "compileKotlin"(KotlinCompile::class) {
+                    dependsOn(generateImplicitImportsFile)
+                    kotlinOptions {
+                        freeCompilerArgs += listOf(
+                            "-script-templates", scriptTemplates,
+                            // Propagate implicit imports and other settings
+                            "-Xscript-resolver-environment=kotlinDslImplicitImportsFile=\"${implicitImportsFile.get().asFile.canonicalPath}\""
```
eskatos: :x: shouldn't the implicit import file be explicitly registered as a file input for the `KotlinCompile` task? we should also prove the changes in this PR don't break the cacheability of `KotlinCompile`

Issue #754, opened 2018-03-14
Use metadata sources API to define the embedded-kotlin-repo as artifact-only repository


jjohannes:
There is no need to check for metadata files, as they never exist. Also we will no longer check for `jars` at all in Gradle 5.0 which will break the repository (can be tested already with `enableFeaturePreview('GRADLE_METADATA')`).

The fix is to add
```
  repo.metadataSources {
    it.artifact()
  }
```

here
https://github.com/gradle/kotlin-dsl/blob/develop/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/EmbeddedKotlinProvider.kt#L85-L88

bamboo:
Closing as delivered in b70ab790770752297a48eddffc585d8101111d2b.

PR #759, opened 2018-03-15
Fix Gradle sources resolution in face of -bin snapshot distribution with GRADLE_METADATA feature preview enabled

eskatos:
Before this change,
with a snapshot Gradle distro `-bin`,
and `enableFeaturePreview("GRADLE_METADATA")` in `settings.gradle.kts`,
when opening a script in IntelliJ,
the script dependencies resolver throws: https://gist.github.com/eskatos/170932f5aeaf92495d6ae19e89bbe6ac,
no dependencies are returned to IntelliJ,
which highlight the whole file red.

This PR fixes the synthetic repository used to resolve Gradle sources to declare it contains artifacts only. This is basically the same fix as for #754.

I confirmed manually that it fixes the issue at hand with `enableFeaturePreview("GRADLE_METADATA")`.

### DIFF
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/SourceDistributionProvider.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/SourceDistributionProvider.kt
@@ -92,6 +92,9 @@ class SourceDistributionResolver(val project: Project) : SourceDistributionProvi
         val repoName = repositoryNameFor(gradleVersion)
         name = "Gradle $repoName"
         setUrl("https://services.gradle.org/$repoName")
+        metadataSources { sources ->
+            sources.artifact()
+        }
         layout("pattern") {
             val layout = it as IvyPatternRepositoryLayout
             if (isSnapshot(gradleVersion)) {


provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/SourceDistributionProvider.kt
```diff
@@ -45,6 +45,15 @@ class SourceDistributionResolver(val project: Project) : SourceDistributionProvi
     }
 
     override fun sourceDirs(): Collection<File> =
+        try {
+            collectSourceDirs()
+        } catch (e: Exception) {
+            e.printStackTrace()
```
melix: Is this intentional? Shouldn't it use a logger?

eskatos: Good point.

This code will run inside IntelliJ in the context of the script dependencies resolver. So `ex.printStackTrace()` will go into the void. We should report this exception up for it to be logged into the script dependencies resolver logs so we can diagnose issues properly.

I'll remove the leniency aspect from this PR so it contains the GRADLE_METADATA fix only. And follow up with another leniency PR that propagates the error up for logging.

eskatos: Actually ... you can scratch the above, I mixed up my mind.

This code will run in a Gradle daemon behind the TAPI builders that the script dependencies resolver will invoke. So it should be easy enough to use a logger to get the traces into the daemon logs.

Issue #611, opened 2017-11-26
Invalid code generated for accessor of extension with generic collection element type


bamboo:
The `konan` plugin from the [kotlin-native project](https://github.com/JetBrains/kotlin-native) registers the `konanArtifacts` extension for which `kotlin-dsl` generates the following accessor:

```kotlin
val Project.`konanArtifacts`: org.gradle.api.NamedDomainObjectContainer<org.jetbrains.kotlin.gradle.plugin.KonanBuildingConfig> = ...
```

The generated accessor fails to compile for `KonanBuildingConfig` is missing a generic argument:

```
e: src/org/gradle/kotlin/dsl/accessors.kt:68:108: One type argument expected for class KonanBuildingConfig<T : KonanBuildingTask>
```

This behavior can be reproduced with the following build scripts:

```kotlin
// settings.gradle.kts
pluginManagement {
    repositories {
        maven(url = "https://dl.bintray.com/jetbrains/kotlin-native-dependencies")
    }
    resolutionStrategy {
        eachPlugin {
            if (requested.id.id == "konan") {
                useModule("org.jetbrains.kotlin:kotlin-native-gradle-plugin:0.4")
            }
        }
    }
}

// build.gradle.kts
plugins {
    id("konan")
}
```

bamboo:
Upstream PR of interest: JetBrains/kotlin-native#1088

eskatos:
Consider `<*>` for when it is not possible.

eskatos:
Closing as delivered in 54d5591

PR #765, opened 2018-03-16
 Generate compilable accessors in face of type erasure

eskatos:
See #611


provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/TypeAccessibilityProviderTest.kt
```diff
@@ -21,6 +21,30 @@ class TypeAccessibilityProviderTest : TestWithClassPath() {
...
+
+        classNamesFromTypeString("java.lang.String").apply {
+            assertThat(all, hasItems("java.lang.String"))
+            assertThat(leafs, hasItems("java.lang.String"))
+        }
+
+        classNamesFromTypeString("java.util.Map<java.util.List, java.util.Set>").apply {
+            assertThat(all, hasItems("java.util.Map", "java.util.List", "java.util.Set"))
+            assertThat(leafs, hasItems("java.util.List", "java.util.Set"))
+        }
```
JLLeitschuh: :o:
Test for:
`java.util.Map<java.util.List<String>, java.util.Set<String>>` or is that not relevant? Alsom what about nullable types?

eskatos: `<String>` would be ignored as it is a primitive
No nullability handled in this context.

Issue #319, opened 2017-03-31
`configure<ExtensionType> { ... }` and `the<ExtensionType>()` allows generic `ExtensionType`


bamboo:
Given an extension [registered under the generic public type](https://github.com/gradle/gradle/blob/408b10a601f95542dbd015f906f197cdd20b1a3f/subprojects/core/src/main/java/org/gradle/api/plugins/ExtensionContainer.java#L67-L67) `NamedDomainObjectContainer<SomeType>`, the following code will configure it:

```kotlin
configure<NamedDomainObjectContainer<SomeType>> {
  val someValue by creating {
    someProperty = 42
  }
}
```

Using [`ExtensionContainer#findByType<T>(TypeOf<T>)`](https://github.com/gradle/gradle/blob/408b10a601f95542dbd015f906f197cdd20b1a3f/subprojects/core/src/main/java/org/gradle/api/plugins/ExtensionContainer.java#L185-L185) under the hood.

eskatos:
Related to #611

eskatos:
Closing as delivered in 245aaea

PR #766, opened 2018-03-19
Let the<T>() and configure<T>() support nested generic queries

eskatos:
by using `TypeOf` type tokens.

See #319


provider/src/test/kotlin/org/gradle/kotlin/dsl/ProjectExtensionsTest.kt
```diff
@@ -46,4 +53,144 @@ class ProjectExtensionsTest {
         val p by project
         assertThat(p, equalTo<Any>(42))
     }
+
+    @Test
+    fun `can get generic project extension by type`() {
+
+        val project = mock<Project>()
+        val convention = mock<Convention>()
+        val extension = mock<NamedDomainObjectContainer<List<String>>>()
```
JLLeitschuh: Does mockito support nested generics like this correctly?

provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/ProjectSchemaTest.kt
```diff
@@ -2,6 +2,8 @@ package org.gradle.kotlin.dsl.accessors
 
 import org.gradle.internal.classpath.ClassPath
 
+import org.gradle.kotlin.dsl.typeOf
```
JLLeitschuh: :o: Unnecessary addition?

provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/TypeAccessibilityProviderTest.kt
```diff
@@ -2,6 +2,8 @@ package org.gradle.kotlin.dsl.accessors
 
 import org.gradle.internal.classpath.ClassPath
 
+import org.gradle.kotlin.dsl.typeOf
```
JLLeitschuh: :o: Unnecessary addition?

provider/src/main/kotlin/org/gradle/kotlin/dsl/TypeOfExtensions.kt
```diff
@@ -0,0 +1,28 @@
...
+
+
+/**
+ * Creates an instance of [TypeOf] for the given parameterized type.
+ *
+ * @param T the type
+ * @return the [TypeOf] that captures the generic type of the given parameterized type
+ */
+inline fun <reified T> typeOf() =
+    object : TypeOf<T>() {}
```
JLLeitschuh: Love this. Is this intended for consumers though. Just making sure.
Seems like this trick is pretty common in libraries that need to save the reflected type.
https://github.com/JLLeitschuh/kotlin-guiced/blob/master/core/src/main/kotlin/org/jlleitschuh/guice/Key.kt#L5

Curiosity though, how would this handle nullable types?
Eg:
`typeOf<String?>`?

provider/src/test/kotlin/org/gradle/kotlin/dsl/ProjectExtensionsTest.kt
```diff
@@ -46,4 +53,144 @@ class ProjectExtensionsTest {
...
+        val conventionType = typeOf<JavaPluginConvention>()
+
+        whenever(project.convention)
+            .thenReturn(convention)
+        whenever(convention.getByType(eq(conventionType)))
+            .thenThrow(UnknownDomainObjectException::class.java)
+
+        try {
+            project.the<JavaPluginConvention>()
+            fail("UnknownDomainObjectException not thrown")
```
JLLeitschuh: 👍

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
```diff
@@ -826,6 +826,68 @@ class GradleKotlinDslIntegrationTest : AbstractIntegrationTest() {
...
+                put("zero", 0)
+            }
+            require(the<MutableMap<String, Int>>() == mapOf("deep" to 42, "zero" to 0))
+
+            require(the<MutableMap<*, *>>() == mapOf("foo" to "bar", "bazar" to "cathedral"))
+
+            configure<NamedDomainObjectContainer<my.Book>> {
+                create("The Dosadi experiment")
+            }
+            require(the<NamedDomainObjectContainer<my.Book>>().size == 1)
```
JLLeitschuh: What is this API? Where is it coming from? I've never seen it before, but I like it.

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
```diff
@@ -826,6 +826,68 @@ class GradleKotlinDslIntegrationTest : AbstractIntegrationTest() {
...
+            }
+        """)
+
+        withFile("buildSrc/src/main/kotlin/my/MyPlugin.kt", """
+            package my
+
+            import org.gradle.api.*
+            import org.gradle.kotlin.dsl.*
+
+            open class Book(val name: String)
```
JLLeitschuh: Does this need to be open to work?

provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/ProjectSchemaTest.kt
eskatos: Nope, it moved.
See e16e0c6d2b07dd964293fa20246c3c8a80c40e2c

provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/TypeAccessibilityProviderTest.kt
eskatos: Nope, it moved.
See e16e0c6d2b07dd964293fa20246c3c8a80c40e2c

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
eskatos: It's `kotlin-stdlib`: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/require.html

provider/src/main/kotlin/org/gradle/kotlin/dsl/TypeOfExtensions.kt
eskatos: `TypeOf` is Gradle public API, so yes, intended for consumers.

Good point, it wouldn't handle Kotlin nullability. So it should be `<reified T : Any>` in order to limit usage to non-nullable type parameters only.

provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/TypeAccessibilityProviderTest.kt
JLLeitschuh: Okay.

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
JLLeitschuh: Cool!

provider/src/main/kotlin/org/gradle/kotlin/dsl/TypeOfExtensions.kt
eskatos: After thinking about this a bit more, `typeOf<T>()` should not limit usage to non-nullable types only. `the<T>()` and `configure<T>()` do because they throw if the underlying object cannot be found.

Other consumers of `typeOf<T>()` may pass nullable types. It just happens that nullability is ignored by the underlying java implementations. And nothing stops one to write `typeOf<NonNullable<Nullable?>>()` anyway.

JLLeitschuh: Fair enough.

Issue #669, opened 2018-01-17
Automatically expose `*.gradle.kts` from regular Kotlin source sets as binary plugins to `java-gradle-plugin` (binary kts plugins)


bamboo:
- [x] project script plugins from the plugin sourceset
- [x] settings script plugins from the plugin sourceset
- [x] init script plugins from the plugin sourceset
- [x] infer plugin id from script file name and package name

PR #767, opened 2018-03-19
Expose precompiled scripts as Gradle plugins

bamboo:
The Gradle plugin id is derived from the script file name and its package declaration if any.

The implementation is currently limited to scripts under `src/main/kotlin`. Future commits will lift this restriction.

See #669


plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/precompiled/ScriptPlugin.kt
```diff
@@ -0,0 +1,122 @@
...
+}
+
+
+private
+fun packageNameOf(file: File): String? =
+    packageNameOf(file.readText())
+
+
+private
+fun packageNameOf(code: String): String? =
```
eskatos: ❓  given this will happen before `:kotlinCompile` I wonder about the behavior if the lexer throws.

plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/precompiled/PrecompiledScriptPlugins.kt
```diff
@@ -28,51 +32,168 @@ import org.gradle.kotlin.dsl.precompile.PrecompiledSettingsScript
...
+
+private
+fun Project.implicitImports(): List<String> =
+    serviceOf<ImplicitImports>().list
+
+
+private
+fun Project.exposeScriptsAsGradlePlugins() {
+
+    // TODO: script plugins from custom source-set dirs
```
eskatos: I'd be fine with handling a single source set for now.
In that case consider using `gradlePlugin.pluginSourceSet`.

plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/precompiled/PrecompiledScriptPlugins.kt
```diff
@@ -28,51 +32,168 @@ import org.gradle.kotlin.dsl.precompile.PrecompiledSettingsScript
...
+
+        getByName("compileKotlin") {
+            it.dependsOn(generateScriptPluginAdapters)
+        }
+    }
+}
+
+
+internal
+fun ScriptPlugin.writeScriptPluginAdaterTo(outputDir: File) =
```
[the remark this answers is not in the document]
bamboo: 👍

plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/precompiled/ScriptPlugin.kt
```diff
@@ -0,0 +1,122 @@
...
+        while (tokenType == KtTokens.IDENTIFIER || tokenType == KtTokens.DOT) {
+            append(tokenText)
+            advance()
+        }
+        toString()
+    }
+
+
+private
+fun KotlinLexer.skipWhiteSpaceAndComments() {
```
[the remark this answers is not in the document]
bamboo: I didn't want to couple them.

plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/precompiled/PrecompiledScriptPlugins.kt
bamboo: Makes a lot of sense. Done.

plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/precompiled/ScriptPlugin.kt
bamboo: I would expect the build to fail.

eskatos: Right. I'm wondering *how* it would fail and if the error would make sense to the user
I think we should have a test for this and make sure we display a reasonable error.

PR #782, opened 2018-03-23
Allow non-strict Gradle classloading to support evaluationDependsOn() & COD

eskatos:
This should address both #311 and #607

### DIFF
--- a/provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
+++ b/provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
@@ -5,6 +5,8 @@ import okhttp3.mockwebserver.MockWebServer
 
 import org.gradle.api.JavaVersion
 
+import org.gradle.api.internal.initialization.DefaultClassLoaderScope
+
 import org.gradle.kotlin.dsl.embeddedKotlinVersion
 import org.gradle.kotlin.dsl.fixtures.AbstractIntegrationTest
 import org.gradle.kotlin.dsl.fixtures.DeepThought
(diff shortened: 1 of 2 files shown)


provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/GradleKotlinDslIntegrationTest.kt
```diff
@@ -894,6 +896,28 @@ class GradleKotlinDslIntegrationTest : AbstractIntegrationTest() {
...
     }
 
+    /**
+     * Assert that the Kotlin DSL provider don't force strict classloading
+     * by exercising a case where :b class loader scopes aren't locked when evaluating :b:c.
+     */
+    @Test
+    fun `allow for non-strict Gradle classloading`() {
+
+        val previous = System.getProperty(DefaultClassLoaderScope.STRICT_MODE_PROPERTY)
```
bamboo: :x: `withSystemProperty(STRICT_MODE_PROPERTY, "false") { ...`

oehme:
I agree we should fix this upstream. The Kotlin DSL shouldn't make that decision though, which is why I'm 👍 on this change.

bamboo:
I understand where you're coming from @oehme but I still don't think the decision the Kotlin DSL is making here - `require(scope.isLocked)` - is unreasonable at this level.

It's there to enforce the reproducibility of computing the classpath of a project.

The alternative, to let the classpath vary depending on the order the projects are evaluated, could lead to hard to debug script compilation errors particularly when combined with configure-on-demand:

> Why my script fails to compile when I run `./gradlew :foo:bar` but compiles cleanly when I run `./gradlew bar`?

Or am I overstating the risk?

oehme:
The same is true for Groovy scripts and has been for a long time. That's why I don't think this check belongs in the Kotlin DSL. We should fix this centrally.

eskatos:
> The same is true for Groovy scripts and has been for a long time. That's why I don't think this check belongs in the Kotlin DSL. We should fix this centrally.

@oehme could you please express your concern an an issue on `gradle/gradle` so we can refer to it if need be? or there is one I couldn't find already?

eskatos:
Closing as superseded by https://github.com/gradle/gradle/issues/4823

PR #786, opened 2018-03-24
Let source roots of multi-project buildSrc be available to the IDE

eskatos:


provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/SourcePathProvider.kt
```diff
@@ -40,13 +43,18 @@ object SourcePathProvider {
...
     /**
-     * Returns all conventional source directories under buildSrc if any.
-     *
-     * This won't work for buildSrc projects with a custom source directory layout
-     * but should account for the majority of cases and it's cheap.
+     * Returns source directories from buildSrc if any.
      */
     private
     fun buildSrcRootsOf(projectRoot: File): Collection<File> =
+        projectRoot.resolve("buildSrc/build/$buildSrcSourceRootsFilePath")
```
bamboo: :x: The task is writing to `layout.buildDirectory.file(buildSrcSourceRootsFilePath)` and the provider is reading from `projectRoot.resolve("buildSrc/build/$buildSrcSourceRootsFilePath")` which wouldn't match for custom layouts. Hardcoding `build` on both sides seems slightly better:
1. The IDE experience would not deteriorate given a custom `layout.buildDirectory` in `buildSrc`
2. We are more likely to get accurate bug reports when/if this persistent `buildSrc/build` directory becomes a problem than for IDE features seemingly not working

As an alternative, have you considered embedding the `source-roots.txt` file as a resource?

tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/BuildSrcSourceRootsConfigurationAction.kt
```diff
@@ -0,0 +1,109 @@
...
+                    destinationFile.set(layout.buildDirectory.file(buildSrcSourceRootsFilePath))
+                }
+                getByName("jar").finalizedBy(generateSourceRoots)
+            }
+        }
+    }
+
+    private
+    fun Project.projectDependenciesSourceRootsFrom(configurationName: String) = provider {
+        configurations.getByName(configurationName).incoming.resolutionResult.allComponents.asSequence()
```
bamboo: Very nice! 👍

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/ScriptModelIntegrationTest.kt
```diff
@@ -72,10 +72,74 @@ abstract class ScriptModelIntegrationTest : AbstractIntegrationTest() {
...
+    fun withMainSourceSetJavaKotlinIn(projectDir: String) =
+        ProjectSourceRoots(existing(projectDir), listOf("main"), listOf("java", "kotlin"))
+
+    protected
+    fun matchesProjectsSourceRoots(vararg projectSourceRoots: ProjectSourceRoots): Matcher<Iterable<File>> {
+
+        fun hasLanguageDir(base: File, set: String, lang: String): Matcher<Iterable<*>> =
+            hasItem(base.resolve("src/$set/$lang"))
+
+        val all = mutableListOf<Matcher<Iterable<*>>>()
```
bamboo: :x: It might not be terribly efficient but I would go with the collection pipeline version:

```kotlin

    protected
    fun matchesProjectsSourceRoots(vararg projectSourceRoots: ProjectSourceRoots): Matcher<Iterable<File>> {

        fun hasLanguageDir(base: File, set: String, lang: String): Matcher<Iterable<*>> =
            hasItem(base.resolve("src/$set/$lang"))

        return allOf(
            *projectSourceRoots
                .filter { it.languages.isNotEmpty() }
                .flatMap { sourceRoots ->
                    val languageDirs =
                        sourceRoots.sourceSets.flatMap { sourceSet ->
                            listOf("java", "kotlin").map { language ->
                                val hasLanguageDir = hasLanguageDir(sourceRoots.projectDir, sourceSet, language)
                                if (language in sourceRoots.languages) hasLanguageDir
                                else not(hasLanguageDir)
                            }
                        }

                    val resourceDirs =
                        sourceRoots.sourceSets.map { sourceSet ->
                            hasLanguageDir(sourceRoots.projectDir, sourceSet, "resources")
                        }

                    languageDirs + resourceDirs
                }.toTypedArray())
    }
```

provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/SourcePathProvider.kt
eskatos: I did initially embed the `source-roots.txt` file as a resource in `buildSrc.jar` but given we don't generate that file on the regular usage, only when in classpath mode, it would have caused too much classpath invalidation. So, I went back to directly pointing at the file.

Otherwise, good point about the `build` path segment, I did what you suggested.

eskatos:
@bamboo ready for another round

provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/SourcePathProvider.kt
bamboo: What if it is always generated with a cacheable task?

eskatos: This would work. But it would mean adding the project configuration and task creation done by `BuildSrcSourceRootsConfigurationAction` plus `projectDependenciesSourceRootsFrom()` overhead to all builds out there. Execution of `projectDependenciesSourceRootsFrom()` would be required always even when fetching from cache as it's actually the inputs of the task.

bamboo: Makes sense.

tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/BuildSrcSourceRootsConfigurationAction.kt
```diff
@@ -0,0 +1,109 @@
...
+            .toList()
+    }
+
+    private
+    fun Sequence<ResolvedComponentResult>.projectDependenciesIdentifiers() =
+        mapNotNull { it.id as? ProjectComponentIdentifier }
+
+    private
+    fun Sequence<Project>.withJavaBasePlugin() =
+        filter { it.plugins.hasPlugin((JavaBasePlugin::class.java)) }
```
JLLeitschuh: :x: Unnecessary double parentheses.

PR #794, opened 2018-03-26
Provide correct classpath for precompiled script plugins to the IDE

bamboo:
The classpath of a precompiled script plugin is the compile classpath of its enclosing source-set.

This PR doesn't change the source path behaviour which will be improved in a separate PR.

### DIFF
--- a/provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/PrecompiledScriptPluginModelIntegrationTest.kt
+++ b/provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/PrecompiledScriptPluginModelIntegrationTest.kt
@@ -13,8 +13,8 @@ class PrecompiledScriptPluginModelIntegrationTest : ScriptModelIntegrationTest()
     @Test
     fun `given a single project build, the classpath of a precompiled script plugin is the compile classpath of its enclosing source-set`() {
 
-        val compileDependency =
-            withFile("compile.jar")
+        val implementationDependency =
+            withFile("implementation.jar")
 
         val classpathDependency =
             withFile("classpath.jar")
@@ -31,7 +31,7 @@ class PrecompiledScriptPluginModelIntegrationTest : ScriptModelIntegrationTest()
             }
 
             dependencies {
-                compile(files("${compileDependency.name}"))
+                implementation(files("${implementationDependency.name}"))
             }
         """)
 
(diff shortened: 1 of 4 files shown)


tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinBuildScriptModelBuilder.kt
```diff
@@ -105,6 +117,35 @@ object KotlinBuildScriptModelBuilder : ToolingModelBuilder {
...
+private
+fun findSourceSetOfFileIn(projects: Iterable<Project>, file: File): SourceSet? =
+    projects
+        .asSequence()
+        .mapNotNull { it.findSourceSetOf(file) }
+        .firstOrNull()
+
+
+private
+fun Project.findSourceSetOf(scriptFile: File): SourceSet? =
```
eskatos: 👍

provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/PrecompiledScriptPluginModelIntegrationTest.kt
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            buildscript {
+                dependencies {
+                    classpath(files("${classpathDependency.name}"))
+                }
+            }
+
+            dependencies {
+                compile(files("${compileDependency.name}"))
```
eskatos: :x: should use `implementation` instead

tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinBuildScriptModelBuilder.kt
```diff
@@ -105,6 +117,35 @@ object KotlinBuildScriptModelBuilder : ToolingModelBuilder {
...
+        it.allSource.contains(scriptFile)
+    }
+
+
+private
+fun precompiledScriptPluginModelBuilder(enclosingSourceSet: SourceSet, modelRequestProject: Project): KotlinScriptTargetModelBuilder =
+    KotlinScriptTargetModelBuilder(
+        project = modelRequestProject,
+        scriptClassPath = DefaultClassPath(enclosingSourceSet.compileClasspath),
+        sourceLookupScriptHandlers = emptyList())
```
eskatos: :x: unnecessary parameter, it has the same default value

PR #798, opened 2018-03-28
Normalize `inline` modifier layout

bamboo:


plugins-experiments/src/test/kotlin/org/gradle/kotlin/dsl/experiments/plugins/GradleKotlinDslKtlintConventionPluginTest.kt
```diff
@@ -79,8 +79,8 @@ class GradleKotlinDslKtlintConventionPluginTest : AbstractPluginTest() {
                 private lateinit
                 var description: String
 
-                private
-                inline fun something() = Unit
+                private inline
+                fun something() = Unit
```
JLLeitschuh: Why isn't this just one line?

bamboo: @JLLeitschuh that's a test case proving our ktlint rules will complain when `private` and `inline` are found in the same line.

Issue #797, opened 2018-03-27
Dependency on Configuration throws IllegalStateException: add("conf", dependencyNotation) must not be null


eskatos:
```kotlin
plugins { java }
configurations.create("some")
dependencies {
    "testRuntime"(configurations["some"])
}
```

fails with `add("testRuntime", dependencyNotation) must not be null`
https://scans.gradle.com/s/vkvou6fblhbkc

This happens because:
- [DependencyHandlerScope { String.invoke(dep: Any) }](https://github.com/gradle/kotlin-dsl/blob/2e5ca196e63e0f76cb85835fd177b6ff4ab5e468/provider/src/main/kotlin/org/gradle/kotlin/dsl/DependencyHandlerScope.kt#L40) doesn't allow a nullable return type
- the actual [implementation](https://github.com/gradle/gradle/blob/7cb91e497d257049b4a36367e3e41aa08ef2d92e/subprojects/dependency-management/src/main/java/org/gradle/api/internal/artifacts/dsl/dependencies/DefaultDependencyHandler.java#L103-L110) returns `null` when passed a `Configuration`

This got unnoticed until now mainly because [DependencyHandler.add()](https://github.com/gradle/gradle/blob/7cb91e497d257049b4a36367e3e41aa08ef2d92e/subprojects/core-api/src/main/java/org/gradle/api/artifacts/dsl/DependencyHandler.java#L261) is not annotated with `@Nullable`.

bamboo:
Closed by accident. Still need to handle the merge on the `gradle/gradle` side.

PR #809, opened 2018-04-03
Allow adding a dependency on a `Configuration` instance

bamboo:
By taking into account that `DependencyHandler.add(String, Any)` will
return `null` when given a `Configuration`.

See #797


bamboo:
@eskatos Of course! Done.

JLLeitschuh:
:o: Wouldn't it be better to provide an overload that takes a `Configuration` and only have that one return null? That way in the general case, you don't have to null check?

PR #815, opened 2018-04-09
[WIP] Spike - Generate Kotlin extensions for Gradle API

eskatos:
This PR is a spike on generating Kotlin extensions over the Gradle API that:
- replace `Class<T>` parameters with `KClass<T>`
- reify type argument of `TypeOf` or `Class` taking methods, favoring `TypeOf<T>` as it allows to capture type arguments fully

For:

```java
public interface ObjectFactory {
    <T extends Named> T named(Class<T> type, String name);
}
```

it generates the following (comments and annotations ommited):

```kotlin
fun <T : Named> ObjectFactory.named(p0: KClass<T>, p1: String): T =
    named(p0.java, p1)

inline fun <reified T : Named> ObjectFactory.named(p1: String): T =
    named(T::class.java, p1)
```

It also maps arrays of `Class` to Kotlin varargs, e.g.:

```kotlin
fun ArtifactResolutionQuery.withArtifacts(p0: KClass<Component>, vararg p1: KClass<Artifact>): ArtifactResolutionQuery =
    withArtifacts(p0.java, *p1.map { it.java }.toTypedArray())
```

And translates `Action<T>` to `T.() -> Unit`, e.g.:

```kotlin
inline fun <reified T : Any> ExtensionContainer.configure(noinline p1: T.() -> Unit): Unit =
    configure(typeOf<T>(), p1)
```

Unfortunately, parameter names aren't present in the bytecode so extensions use numbered parameter names. Once Gradle requires Java 8 and is compiled with the `-parameters` flag we'll be able to read parameter names from the bytecode. Client code using named parameters will break when parameter names change.

This process produces extensions that are already present in the Gradle Kotlin DSL api, the latter should be removed. This spike doesn't include the removals, hence using it as this repository's wrapper will fail with conflicting extensions errors. These removals also are a breaking change for client code that uses named parameters.

Just like for extensions for core plugin IDs, the generation happens when generating the `gradle-kotlin-dsl-extensions-XXX.jar`.

On a local cold vm, generation takes <2s, down to 700ms if run repeatedly. Compilation takes ~6/7s, down to ~1.5s if run repeatedly. But this is intended to be run on a cold vm, once per Gradle Kotlin DSL version. We're talking about a total of 8 to 9 seconds.

It goes over the whole Gradle API. This is relatively costly because we need to discover all available types from a ClassPath. Another approach could be to generate extensions only for types from the implicit imports list, thus removing the type discovery cost.

We could win, say, a second, by e.g. using the implicit imports as an input and parallelizing the generation. Some other optimizations are probably lurking here and there.  But, Kotlin compilation largely dominates.

### DIFF
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
@@ -15,6 +15,8 @@
  */
 package org.gradle.kotlin.dsl.codegen
 
+import org.gradle.api.Incubating
+
 import org.gradle.kotlin.dsl.accessors.primitiveTypeStrings
 import org.gradle.kotlin.dsl.support.ClassBytesRepository
 import org.gradle.kotlin.dsl.support.classPathBytesRepositoryFor
@@ -39,6 +41,9 @@ import org.jetbrains.org.objectweb.asm.tree.MethodNode
 
 import java.io.Closeable
 import java.io.File
+import java.util.Objects
+
+import javax.annotation.Nullable
 
 import kotlin.LazyThreadSafetyMode.NONE
 
@@ -80,28 +85,29 @@ class ApiTypeProvider(private val repository: ClassBytesRepository) : Closeable 
 
     fun type(sourceName: String): ApiType? = open {
         apiTypesBySourceName.computeIfAbsent(sourceName) {
-            repository.classBytesFor(sourceName)?.let { apiTypeFor(sourceName, it) }
+            repository.classBytesFor(sourceName)?.let { apiTypeFor(sourceName, { it }) }
         }?.invoke()
     }
 
     fun allTypes(): Sequence<ApiType> = open {
-        repository.allClassesBytesBySourceName().mapNotNull { (sourceName, classBytes) ->
+        repository.allClassesBytesBySourceName().map { (sourceName, classBytes) ->
             apiTypesBySourceName.computeIfAbsent(sourceName) {
-                apiTypeFor(sourceName, classBytes())
-            }
+                apiTypeFor(sourceName, classBytes)
+            }!!
         }.map { it() }
     }
 
     private
-    fun apiTypeFor(sourceName: String, classBytes: ByteArray) = {
+    fun apiTypeFor(sourceName: String, classBytes: () -> ByteArray) = {
         ApiType(sourceName, classNodeFor(classBytes), { type(it) })
     }
 
     private
-    fun classNodeFor(classBytes: ByteArray) =
+    fun classNodeFor(classBytesSupplier: () -> ByteArray) = {
         ApiTypeClassNode().also {
-            ClassReader(classBytes).accept(it, SKIP_DEBUG or SKIP_CODE or SKIP_FRAMES)
+            ClassReader(classBytesSupplier()).accept(it, SKIP_DEBUG or SKIP_CODE or SKIP_FRAMES)
         }
+    }
 
     private
     fun <T> open(action: () -> T): T =
@@ -111,27 +117,32 @@ class ApiTypeProvider(private val repository: ClassBytesRepository) : Closeable 
 
 
 internal
-class ApiType(
+data class ApiType(
     val sourceName: String,
-    private val delegate: ClassNode,
+    private val delegateSupplier: () -> ClassNode,
     private val typeIndex: ApiTypeIndex
 ) {
 
-    val isPublic: Boolean =
-        (ACC_PUBLIC and delegate.access) > 0
+    val isPublic: Boolean
+        get() = (ACC_PUBLIC and delegate.access) > 0
 
     val isDeprecated: Boolean
-        get() = delegate.visibleAnnotations.hasDeprecated
+        get() = delegate.visibleAnnotations.has<java.lang.Deprecated>()
 
     val isIncubating: Boolean
-        get() = delegate.visibleAnnotations.hasIncubating
+        get() = delegate.visibleAnnotations.has<Incubating>()
 
-    val formalTypeParameters: List<ApiTypeUsage> by lazy(NONE) {
-        visitedSignature?.formalTypeParameterUsages(typeIndex) ?: emptyList()
+    val typeParameters: List<ApiTypeUsage> by lazy(NONE) {
+        typeIndex.apiTypeParametersFor(visitedSignature)
     }
 
     val functions: List<ApiFunction> by lazy(NONE) {
         delegate.methods.filter { it.signature != null }.map { ApiFunction(this, it, typeIndex) }
+    }
+
+    private
+    val delegate: ClassNode by lazy(NONE) {
+        delegateSupplier()
     }
 
     private
@@ -140,11 +151,18 @@ class ApiType(
             ClassSignatureVisitor().also { SignatureReader(signature).accept(it) }
         }
     }
+
+    override fun equals(other: Any?): Boolean =
+        if (other !is ApiType) false
+        else Objects.equals(sourceName, other.sourceName)
+
+    override fun hashCode(): Int =
+        Objects.hash(sourceName)
 }
 
 
 internal
-class ApiFunction(
+data class ApiFunction(
     private val owner: ApiType,
     private val delegate: MethodNode,
     private val typeIndex: ApiTypeIndex
@@ -157,34 +175,24 @@ class ApiFunction(
         (ACC_PUBLIC and delegate.access) > 0
 
     val isDeprecated: Boolean
-        get() = owner.isDeprecated || delegate.visibleAnnotations.hasDeprecated
+        get() = owner.isDeprecated || delegate.visibleAnnotations.has<java.lang.Deprecated>()
 
     val isIncubating: Boolean
-        get() = owner.isIncubating || delegate.visibleAnnotations.hasIncubating
+        get() = owner.isIncubating || delegate.visibleAnnotations.has<Incubating>()
 
     val isStatic: Boolean =
         (ACC_STATIC and delegate.access) > 0
 
-    val formalTypeParameters: List<ApiTypeUsage> by lazy(NONE) {
-        visitedSignature?.formalTypeParameterUsages(typeIndex)
-            ?: emptyList()
+    val typeParameters: List<ApiTypeUsage> by lazy(NONE) {
+        typeIndex.apiTypeParametersFor(visitedSignature)
     }
 
     val parameters: List<ApiFunctionParameter> by lazy(NONE) {
-        delegate.visibleParameterAnnotations?.map { it.hasNullable }.let { parametersNullability ->
-            visitedSignature?.functionParameters(typeIndex, parametersNullability)
-                ?: Type.getArgumentTypes(delegate.desc).mapIndexed { idx, p ->
-                    val isNullable = parametersNullability?.get(idx) == true
-                    apiFunctionParameter(idx, typeIndex.apiTypeUsage(p.className, isNullable))
-                }
-        }
+        typeIndex.apiFunctionParametersFor(delegate, visitedSignature)
     }
 
     val returnType: ApiTypeUsage by lazy(NONE) {
-        delegate.visibleAnnotations.hasNullable.let { isNullable ->
-            visitedSignature?.returnType(typeIndex, isNullable)
-                ?: typeIndex.apiTypeUsage(Type.getReturnType(delegate.desc).className, isNullable)
-        }
+        typeIndex.apiTypeUsageForReturnType(delegate, visitedSignature?.returnType)
     }
 
     private
@@ -194,49 +202,39 @@ class ApiFunction(
         }
     }
 
-    private
-    val List<AnnotationNode>?.hasNullable: Boolean
-        get() = this?.any { it.desc == "Ljavax/annotation/Nullable;" } ?: false
+    override fun equals(other: Any?): Boolean =
+        if (other !is ApiFunction) false
+        else Objects.equals(owner, other.owner)
+            && Objects.equals(delegate.access, other.delegate.access)
+            && Objects.equals(delegate.name, other.delegate.name)
+            && Objects.equals(delegate.desc, other.delegate.desc)
+            && Objects.equals(delegate.signature, other.delegate.signature)
+
+    override fun hashCode(): Int =
+        Objects.hash(owner, delegate.access, delegate.name, delegate.desc, delegate.signature)
 }
 
 
-private
-val List<AnnotationNode>?.hasDeprecated: Boolean
-    get() = this?.any { it.desc == "Ljava/lang/Deprecated;" } ?: false
-
-
-private
-val List<AnnotationNode>?.hasIncubating: Boolean
-    get() = this?.any { it.desc == "Lorg/gradle/api/Incubating;" } ?: false
-
-
-private
-fun apiFunctionParameter(index: Int, type: ApiTypeUsage) =
-    ApiFunctionParameter("p$index", type)
-
-
 internal
-class ApiFunctionParameter(val name: String, val type: ApiTypeUsage)
-
-
-internal
-class ApiTypeUsage(
+data class ApiTypeUsage(
     val sourceName: String,
     val isNullable: Boolean = false,
     val type: ApiType? = null,
-    val typeParameters: List<ApiTypeUsage> = emptyList(),
-    val bounds: List<ApiTypeUsage> = emptyList()
-) {
-    val isRaw: Boolean
-        get() = typeParameters.isEmpty() && type?.formalTypeParameters?.isEmpty() != false
-}
+    val typeArguments: List<ApiTypeUsage> = emptyList(),
+    val bounds: List<ApiTypeUsage> = emptyList(),
+    val isRaw: Boolean = typeArguments.isEmpty() && type?.typeParameters?.isEmpty() != false
+)
+
+
+internal
+data class ApiFunctionParameter(val name: String, val type: ApiTypeUsage)
 
 
 private
-fun ApiTypeIndex.apiTypeUsage(
+fun ApiTypeIndex.apiTypeUsageFor(
     binaryName: String,
     nullable: Boolean,
-    typeParameterSignatures: List<TypeSignatureVisitor> = emptyList(),
+    typeArgumentsSignatures: List<TypeSignatureVisitor> = emptyList(),
     boundsSignatures: List<TypeSignatureVisitor> = emptyList()
 ): ApiTypeUsage =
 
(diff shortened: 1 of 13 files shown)


provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    private
+    val visitedSignature: ClassSignatureVisitor? by lazy(NONE) {
+        delegate.signature?.let { signature ->
+            ClassSignatureVisitor().also { SignatureReader(signature).accept(it) }
+        }
+    }
+
+    override fun equals(other: Any?): Boolean =
```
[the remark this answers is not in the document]
eskatos: Good point. I reworked this aspect.

provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
```diff
@@ -0,0 +1,431 @@
...
+    val sourceName: String,
+    private val delegateSupplier: () -> ClassNode,
+    private val typeIndex: ApiTypeIndex
+) {
+
+    val isPublic: Boolean
+        get() = (ACC_PUBLIC and delegate.access) > 0
+
+    val isDeprecated: Boolean
+        get() = delegate.visibleAnnotations.has<java.lang.Deprecated>()
```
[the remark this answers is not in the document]
eskatos: This PR only deals with Java bytecode. There would be several other things to do if we want to support Kotlin bytecode, which is not required for now.

provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
```diff
@@ -0,0 +1,431 @@
...
+private
+inline fun <reified T : Any> List<AnnotationNode>?.has() =
+    if (this == null) false
+    else Type.getDescriptor(T::class.java).let { desc -> any { it.desc == desc } }
+
+
+private
+class ApiTypeClassNode : ClassNode(ASM6) {
+
+    override fun visitSource(file: String?, debug: String?) = Unit
```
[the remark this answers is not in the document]
eskatos: To short-circuit some processing done in `ClassNode` we don't care about. This is a performance optimisation.

eskatos:
About the Kotlin compilation performance. After discussing a bit with @bamboo we came to the following: given it is a one-time-per-version thing, we can live with the current timing for a while but this should be optimized at some point. If we generated Kotlin AST and then emitted both source and bytecode we could remove the cost of Kotlin compilation, and probably keep the process duration under 2s. See https://github.com/gradle/kotlin-dsl/issues/817

eskatos:
About the parameter names issue. If we want to tackle it before Gradle distribution JAR `.class` files are compiled with Java 8 we'll have to rely on some other mechanism. [paranamer](https://github.com/paul-hammant/paranamer) comes to mind and should be considered.

JLLeitschuh:
I remember hearing from some of the Spring Devs at Kotlin Conf that they add the Kotlin Meta annotations to their compiled class files to give Kotlin access to those parameter names.

eskatos:
@JLLeitschuh I'm only aware of their usage of `-parameters` with javac 8 and `-java-parameters` with kotlinc, something we'll be able to do once Gradle requires Java 8. Do you happen to remember something different?

build.gradle.kts
```diff
@@ -105,6 +105,11 @@ val copyCurrentDistro by task<Copy> {
     into(customInstallationDir)
     exclude("**/*kotlin*")
 
+    // TODO:pm remove SPIKE
+    from(file("gradle/gradle-api-parameter-names-SPIKE.jar")) {
+        into("lib")
+    }
+
```
eskatos: ❗️This would be generated by https://github.com/gradle/gradle/commit/8d75cc1900ed0f4e3464975e25c5ccae609caf6b

eskatos:
Spike updated with proper parameter names, e.g.:

```kotlin
inline fun <reified T : Any> ExtensionContainer.configure(noinline action: T.() -> Unit): Unit =
    configure(typeOf<T>(), action)

inline fun <reified T : Any> ExtensionContainer.create(name: String, instanceType: KClass<T>, vararg constructionArguments: Any): T =
    create(typeOf<T>(), name, instanceType.java, constructionArguments)

fun ArtifactResolutionQuery.withArtifacts(componentType: KClass<Component>, vararg artifactTypes: KClass<Artifact>): ArtifactResolutionQuery =
    withArtifacts(componentType.java, *artifactTypes.map { it.java }.toTypedArray())
```

JLLeitschuh:
> Do you happen to remember something different?
Not offhand no.

provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
```diff
@@ -0,0 +1,448 @@
...
+
+internal
+typealias ParameterNamesSupplier = (String) -> List<String>?
+
+
+/**
+ * Provides [ApiType] instances by Kotlin source name from a class path.
+ *
+ * Keeps JAR files open for fast lookup, must be closed.
+ * Once closed, type graph navigation from [ApiType] and [ApiFunction] instances will throw.
```
JLLeitschuh: > will throw.....
Seems like an incomplete sentence.

provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
```diff
@@ -0,0 +1,448 @@
...
+    Type.getType(descriptor.toString()).className
+
+
+private
+fun binaryNameOfInternalName(internalName: String): String =
+    Type.getObjectType(internalName).className
+
+
+private
+fun sourceNameOfBinaryName(binaryName: String): String =
```
JLLeitschuh: Is this function taking method names, class names, or fully qualified method names?

provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
```diff
@@ -0,0 +1,448 @@
...
+        "java.util.Iterator" to "kotlin.collections.Iterator",
+        "java.util.Collection" to "kotlin.collections.Collection",
+        "java.util.List" to "kotlin.collections.List",
+        "java.util.ArrayList" to "kotlin.collections.ArrayList",
+        "java.util.Set" to "kotlin.collections.Set",
+        "java.util.HashSet" to "kotlin.collections.HashSet",
+        "java.util.LinkedHashSet" to "kotlin.collections.LinkedHashSet",
+        "java.util.Map" to "kotlin.collections.Map",
+        "java.util.HashMap" to "kotlin.collections.HashMap",
+        "java.util.LinkedHashMap" to "kotlin.collections.LinkedHashMap")
```
JLLeitschuh: Does Kotlin have this mapping somewhere in their compiler that you could use instead?
It may not be part of their public API.
Also, what about the the other collection types like:
`NavigableMap` `SortedSet`, `TreeSet`, ect...?

provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/GradleApiExtensions.kt
```diff
@@ -0,0 +1,438 @@
...
+    val returnType: ApiTypeUsage,
+    val expressionBody: String
+) {
+
+    val signatureKey: Int = Objects.hash(targetType.sourceName, name, parameters.map { it.type })
+
+    fun toKotlinString(): String = StringBuilder().apply {
+        appendln("""
+            /**
+             * $description.
```
JLLeitschuh: Can the `description` contain a `*/` within it?

provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/GradleApiExtensions.kt
```diff
@@ -0,0 +1,438 @@
...
+        """.trimIndent())
+        if (isDeprecated) appendln("""@Deprecated("Deprecated Gradle API")""")
+        if (isIncubating) appendln("@org.gradle.api.Incubating")
+        if (isInline) append("inline ")
+        append("fun ")
+        if (typeParameters.isNotEmpty()) append("${typeParameters.joinInAngleBrackets { (if (it.second) "reified " else "") + it.first.toTypeParameterString() }} ")
+        append(targetType.sourceName)
+        if (targetType.typeParameters.isNotEmpty()) append(targetType.typeParameters.toTypeArgumentsString(targetType))
+        append(".")
+        append(name)
```
JLLeitschuh: Should the name have backticks surrounding it incase it's a keyword like `when`?

provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
eskatos: It is taking binary names of classes, see https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.1

provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
```diff
@@ -0,0 +1,448 @@
...
+        typeParameters[binaryName] = ArrayList(1)
+        currentTypeParameter = binaryName
+    }
+
+    override fun visitClassBound(): SignatureVisitor =
+        TypeSignatureVisitor().also { typeParameters[currentTypeParameter]!!.add(it) }
+
+    override fun visitInterfaceBound(): SignatureVisitor =
+        TypeSignatureVisitor().also { typeParameters[currentTypeParameter]!!.add(it) }
+}
```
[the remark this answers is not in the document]
eskatos: Yes

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/ClassBytesRepository.kt
```diff
@@ -0,0 +1,175 @@
...
+        classBytesSupplierForSourceName(sourceName)?.let { it() }
+
+    /**
+     * All found class files bytes by Kotlin source name.
+     */
+    fun allClassesBytesBySourceName(): Sequence<Pair<String, ClassBytesSupplier>> =
+        classPathFiles.asSequence()
+            .flatMap { sourceNamesFrom(it) }
+            .mapNotNull { sourceName ->
+                classBytesSupplierForSourceName(sourceName)?.let { Pair(sourceName, it) }
```
JLLeitschuh: Bikeshed: `sourceName to it`

provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/GradleApiExtensions.kt
eskatos: There're only two private call sites below, I don't see a need for escaping

provider/src/test/kotlin/org/gradle/kotlin/dsl/codegen/GradleApiExtensionsTest.kt
```diff
@@ -0,0 +1,358 @@
...
+            assertContainsExtension("""
+            @org.gradle.api.Incubating
+            inline fun <reified T : Any> org.gradle.api.model.ObjectFactory.setProperty(): org.gradle.api.provider.SetProperty<T> =
+                setProperty(T::class.java)
+            """)
+
+            assertContainsExtension("""
+            inline fun <reified S : T, T : org.gradle.api.Plugin<*>> org.gradle.api.plugins.PluginCollection<T>.withType(): org.gradle.api.plugins.PluginCollection<S> =
+                withType(S::class.java)
+            """)
```
JLLeitschuh: This perpetuates the "too restrictive" issue, you probably can't do anything about this right now though.
https://github.com/gradle/gradle/issues/4050

provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
eskatos: `NavigableMap`, `SortedSet`, `TreeSet` are not part of `kotlin.collections`, when you use them in Kotlin you the original ones from `java.util.*`

We should at least make sure we conform to https://kotlinlang.org/docs/reference/java-interop.html#mapped-types

provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
JLLeitschuh: Any need to worry about anonymous inner classes generated by the compiler with names like:
`testing/MinimalRequiredCodeTest$simpleExampleCase$1$1`?

Just thinking about a bug I faced with the Junit5 Classloader that had a bug:
https://github.com/gradle/gradle/issues/4921

eskatos: Good point, I added filtering and test coverage for this.

melix:
Nice, I like it!

> We're talking about a total of 8 to 9 seconds.

Would it be possible to pre-compute when we build the distribution? It's 9s added to the API jar generation and Kotlin DSL api jar too. So for a first build, it starts to sum up a lot.

build.gradle.kts
```diff
@@ -105,6 +106,12 @@ val copyCurrentDistro by task<Copy> {
     into(customInstallationDir)
     exclude("**/*kotlin*")
 
+    // TODO:pm remove SPIKE
+    from(file("gradle/gradle-api-parameter-names-SPIKE.jar")) {
+        into("lib")
+        rename { it.replace("SPIKE", GradleVersion.current().baseVersion.version) }
+    }
```
eskatos: ❗️This would be generated by https://github.com/gradle/gradle/commit/15a9e245e55

eskatos:
Kotlin extensions generated for the Gradle public API can be seen in the output of https://builds.gradle.org/repository/download/GradleKotlinDSL_DevelopJava9/12206121:id/reports/provider/build/reports/tests/test/classes/org.gradle.kotlin.dsl.codegen.GradleApiExtensionsTest.html

provider/src/test/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProviderTest.kt
```diff
@@ -102,11 +102,15 @@ class ApiTypeProviderTest : AbstractIntegrationTest() {
...
 
             assertThat(type.functions.filter { it.name == "filter" }.size, equalTo(3))
 
             assertTrue(type.functions.none { it.name == "getRootSpec" })
         }
     }
+
+    private
+    inline fun <reified T> ApiTypeProvider.type() =
+        type(canonicalNameOf<T>())
```
JLLeitschuh: Why not null check the return type here instead of at all of the consumers?

eskatos:
@melix
> Would it be possible to pre-compute when we build the distribution? It's 9s added to the API jar generation and Kotlin DSL api jar too. So for a first build, it starts to sum up a lot.

It's not added to generation of the Kotlin DSL api jar, it is the generation of this very jar. But yes, agreed, it starts to sum up.

It could be achieved with:
- a library that does the generation
- a published Gradle Task for `gradle/gradle` to generate Gradle API extensions at build time
- a service used by `kotlin/provider` to generate Plugin API extensions at runtime

Opened #823

In any case we know of a path to make this fast: #817.

provider/src/test/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProviderTest.kt
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Test
+    fun `includes function overrides that change signature, excludes overrides that don't`() {
+        val jars = listOf(withClassJar("some.jar", AbstractCopyTask::class.java))
+
+        apiTypeProviderFor(jars).use { api ->
+            val type = api.type<AbstractCopyTask>()!!
+
+            assertThat(type.functions.filter { it.name == "filter" }.size, equalTo(3))
```
JLLeitschuh: :o: Looks like a brittle test, that is tied to another overload not being added. I'd suggest using some test type that you can control in this project instead.

provider/src/test/kotlin/org/gradle/kotlin/dsl/codegen/GradleApiExtensionsTest.kt
```diff
@@ -0,0 +1,400 @@
...
+            writeGradleApiExtensionsTo(sourceFile, jars)
+        }.also {
+            println("Generation to file succeeded in ${it}ms")
+        }
+
+        val generatedSource = sourceFile.readText()
+        generatedSource.lines().mapIndexed { idx, line -> "${(idx + 1).toString().padStart(4)}: $line" }.forEach { println(it) }
+
+        assertThat(generatedSource, not(containsString("java.lang.Class")))
+        assertThat(generatedSource, not(containsString("(properties: kotlin.collections.Map<String, *>")))
```
JLLeitschuh: Wouldn't it be
```
not(containsString("(`properties`: kotlin.collections.Map<String, *>"))
```
with the backticks?

eskatos:
After some scrutiny over the generated extensions the following issues need to be fixed.

- Type parameters should be reduced in the presence of reified generics in order for extensions to only expose reified type parameters.
- Reified generic extension target type type parameter should be contravariant (`<in S>`) in order to allow polymorphic usage.
- Extensions for methods taking two `Class` parameters with different type arguments should either be skipped or have each applicable parameter reified.

`DomainObjectCollection`, `BinaryCollection`, `ModelMap` and `BuildController` provide a good sample exercising all the issues described above.

JLLeitschuh:
> Reified generic extension target type type parameter should be contravariant (`<in S>`) in order to allow polymorphic usage.

Don't forget potentially nullable returns so don't bound S. Eg. Don't do `<in S : Any>`

JLLeitschuh:
Actually, on that same point, you should check the generator to make sure that it's correctly handling methods that are annotated such that they could potentially return nullable parameters.

Taking your example from the PR description:

```
public interface ObjectFactory {
    @Nullable
    <T extends Named> T named(Class<T> type, String name);
}
```

Should generate one of the following:
Either this:
```kotlin
fun <T : Named?> ObjectFactory.named(p0: KClass<T>, p1: String): T =
    named(p0.java, p1)

inline fun <reified T : Named?> ObjectFactory.named(p0: String): T =
    named(T::class.java, p0)
```
Or this:
```kotlin
fun <T : Named> ObjectFactory.named(p0: KClass<T>, p1: String): T? =
    named(p0.java, p1)

inline fun <reified T : Named> ObjectFactory.named(p0: String): T? =
    named(T::class.java, p0)
```

Not sure which should be generated, but that needs to be considered.

PR #819, opened 2018-04-13
 Add `buildSrc-plugin` sample

bamboo:


samples/buildSrc-plugin/gradle.properties
```diff
@@ -0,0 +1 @@
+testProp=Hello
```
eskatos: :x: unused file

bamboo: Thanks, Paul.

bamboo:
Thanks, @eskatos. Done.

PR #827, opened 2018-04-18
Process the `buildscript` block of project script plugins

bamboo:


provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptTarget.kt
```diff
@@ -93,7 +93,7 @@ fun projectScriptTarget(
     KotlinScriptTarget(
         host = KotlinScriptHost(project, scriptSource, serviceRegistryOf(project), baseScope, scriptHandler),
         scriptTemplate = KotlinBuildScript::class,
-        buildscriptBlockTemplate = KotlinBuildscriptBlock::class.takeIf { topLevelScript },
+        buildscriptBlockTemplate = KotlinBuildscriptBlock::class,
```
JLLeitschuh: Bahaha. That was pretty simple.

eskatos:
The same should be applied to Settings scripts for consistency.

bamboo:
Good point, Paul, done in e8616ae97e38fa4a2b2559e328c515af282e533f.

PR #832, opened 2018-04-20
Remove usages of TestKit internals

eskatos:
We use the internal `DefaultGradleRunner` in order to set Gradle JVM arguments but they can be set using a gradle.properties file instead.

Triggered by https://github.com/gradle/gradle/issues/5075

### DIFF
--- a/samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/samples.kt
+++ b/samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/samples.kt
@@ -23,11 +23,15 @@ fun copySampleProject(from: File, to: File) {
 
 private
 fun withMergedGradleProperties(gradlePropertiesFile: File, action: () -> Unit) =
-    loadPropertiesFrom(gradlePropertiesFile).also { baseProperties ->
-        gradlePropertiesFile.delete()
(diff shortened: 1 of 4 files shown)


samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/samples.kt
```diff
@@ -11,8 +12,27 @@ val samplesRootDir = File(rootProjectDir, "samples")
...
+            it.deleteRecursively()
+        }
     }
 }
+
+
+private
+fun withMergedGradleProperties(gradlePropertiesFile: File, action: () -> Unit) =
+    loadPropertiesFrom(gradlePropertiesFile).also { baseProperties ->
+        gradlePropertiesFile.delete()
```
bamboo: ❓ The `gradlePropertiesFile.delete()` is confusing here. Maybe inlining the whole method would be clearer?

JLLeitschuh:
Does this work in Debug mode?

JLLeitschuh:
I assumed as much. Is doing this the "recommended" way of configuring the JVM args by test developers?

eskatos:
@JLLeitschuh what about you bring that question up in gradle/gradle#5075?

PR #848, opened 2018-04-28
Kotlin 1.2.41

bamboo:


samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/SamplesSmokeTest.kt
```diff
@@ -39,6 +40,7 @@ class SamplesSmokeTest(
         build("tasks")
     }
 
+    @Ignore("requires distro update")
     @Test
     fun `uses the right Kotlin Gradle Plugin version on `() {
```
JLLeitschuh: This test name seems incomplete.

Issue #343, opened 2017-04-25
Provide convenient configuration for SourceSet extensions and conventions


eskatos:
One frequent example is configuring source set directories.

In Groovy:
```groovy
sourceSets.main.kotlin.srcDirs += 'src/main/myKotlin'
```

In Kotlin today:
```kotlin
java.sourceSets["main"].withConvention(KotlinSourceSet::class) {
    kotlin.srcDir(file("src/main/myKotlin"))
}

```

Desired solution:
```kotlin
java.sourceSets["main"].kotlin {
    kotlin.srcDirs += file("src/main/myKotlin")
}
```

Alternate (working) solution:
```kotlin
val kotlin: KotlinSourceSet by java.sourceSets["main"].extensions
kotlin.kotlin.srcDirs += file("src/main/myKotlin")
```

mkobit:
My current solution is to define something like this in the build script or in `buildSrc`:

```kotlin
val SourceSet.kotlin: SourceDirectorySet
  get() = (this as HasConvention).convention.getPlugin(KotlinSourceSet::class.java).kotlin
```

Then, you can do something that looks close to Groovy:

```kotlin
java.sourceSets["main"].kotlin.srcDirs += file("src/main/myKotlin")
```

Still using the internal `HasConvention` but a somewhat useful way to make the Kotlin source set configuration a little simpler.

bamboo:
@mkobit That property getter definition can be simplified to `withConvention(KotlinSourceSet::class) { kotlin }`

gildor:
But this solution `java.sourceSets["main"].kotlin` still uses convention under the hood as I understand.
Are there any global solutions for such problem or some recommendations for plugin authors?
Maybe make more sense to move `sourceSets` definition from `java.sourceSets` to `kotlin.sourceSets`, so completely rid of convention.

JLLeitschuh:
Related: https://github.com/gradle/gradle/issues/3191

@oehme Had said that the goal was to stop exposing the concept of `convention`? Isn't this a step in the wrong direction?

eskatos:
@gildor, @JLLeitschuh, the reality is that Gradle core plugins do use `Convention`s. And just like with all already existing extensions/conventions support in the Gradle Kotlin DSL, this issue is both about extensions and conventions, preferring extensions.

eskatos:
[...]

PR #850, opened 2018-04-30
Generate accessors for SourceSets extensions and conventions

eskatos:
See #343

### DIFF
--- a/provider-plugins/src/main/kotlin/org/gradle/kotlin/dsl/provider/plugins/DefaultProjectSchemaProvider.kt
+++ b/provider-plugins/src/main/kotlin/org/gradle/kotlin/dsl/provider/plugins/DefaultProjectSchemaProvider.kt
@@ -38,7 +38,7 @@ class DefaultProjectSchemaProvider : ProjectSchemaProvider {
             ProjectSchema(
                 targetSchema.extensions,
                 targetSchema.conventions,
-                accessibleConfigurations(project.configurations.names.toList()))
+                accessibleConfigurationsOf(project))
         }
 }
 
@@ -72,7 +72,7 @@ fun targetSchemaFor(target: Any, targetType: TypeOf<*>): ExtensionConventionSche
             }
         }
         if (target is Project) {
-            target.convention.findPlugin(JavaPluginConvention::class.java)?.sourceSets?.forEach { sourceSet ->
+            sourceSetsOf(target)?.forEach { sourceSet ->
                 collectSchemaOf(sourceSet, typeOfSourceSet)
             }
         }
@@ -95,6 +95,11 @@ fun accessibleConventionsSchema(plugins: Map<String, Any>) =
 
 
 private
+fun sourceSetsOf(project: Project) =
+    project.convention.findPlugin(JavaPluginConvention::class.java)?.sourceSets
+
+
+private
 fun inferPublicTypeOfConvention(instance: Any) =
     if (instance is HasPublicType) instance.publicType
     else TypeOf.typeOf(instance::class.java.firstNonSyntheticOrSelf)
@@ -111,8 +116,8 @@ val Class<*>.firstNonSyntheticOrNull: Class<*>?
 
 
 private
-fun accessibleConfigurations(configurations: List<String>) =
-    configurations.filter(::isPublic)
+fun accessibleConfigurationsOf(project: Project) =
+    project.configurations.names.filter(::isPublic)
 
 
 private
@@ -121,13 +126,11 @@ fun isPublic(name: String): Boolean =
 
 
 private
-val typeOfProject: TypeOf<Project> =
-    typeOf()
+val typeOfProject = typeOf<Project>()
 
 
 private
-val typeOfSourceSet: TypeOf<SourceSet> =
-    typeOf()
+val typeOfSourceSet = typeOf<SourceSet>()
 
 
 internal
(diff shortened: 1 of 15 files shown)


provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/ProjectSchemaAccessorsIntegrationTest.kt
```diff
@@ -436,6 +436,31 @@ class ProjectSchemaAccessorsIntegrationTest : AbstractIntegrationTest() {
...
+                srcDir("another/path")
+            }
+
+            val configured = main.groovy.groovy.srcDirs
+            val expected = linkedSetOf(file("src/main/groovy"), file("some/path"), file("another/path"))
+            require(configured == expected)
+        """)
+
+        build("help")
+    }
```
JLLeitschuh: Should you also add a test for the Kotlin plugin as well?

provider-plugins/src/main/kotlin/org/gradle/kotlin/dsl/provider/plugins/DefaultProjectSchemaProvider.kt
```diff
@@ -0,0 +1,108 @@
...
+                    collectSchemaOf(target.extensions.getByName(schema.name), schema.publicType)
+                }
+            }
+        }
+        if (target is HasConvention) {
+            accessibleConventionsSchema(target.convention.plugins).forEach { name, type ->
+                conventions.add(ProjectSchemaEntry(targetType, name, type))
+                collectSchemaOf(target.convention.plugins[name]!!, type)
+            }
+        }
```
JLLeitschuh: Can something be `ExtensionAware` and `HasConvention`? It might be good to note that in a comment.

provider-plugins/src/main/kotlin/org/gradle/kotlin/dsl/provider/plugins/DefaultProjectSchemaProvider.kt
```diff
@@ -0,0 +1,108 @@
...
+
+
+private
+fun isPublic(name: String): Boolean =
+    !name.startsWith("_")
+
+
+private
+val sourceSetType: TypeOf<SourceSet> =
+    TypeOf.typeOf(SourceSet::class.java)
```
JLLeitschuh: Don't you have an API method somewhere that creates a type of for you from a reified generic?

eskatos: It is not available in this module.

provider-plugins/src/main/kotlin/org/gradle/kotlin/dsl/provider/plugins/DefaultProjectSchemaProvider.kt
eskatos: Yes it can. The if branches a few lines above make it explicit already.

provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/ProjectSchemaAccessorsIntegrationTest.kt
eskatos: It is also already covered in `GenerateProjectSchemaTest`

provider-spi/src/main/kotlin/org/gradle/kotlin/dsl/provider/spi/KotlinTypeStrings.kt
```diff
@@ -0,0 +1,60 @@
...
+    type.run {
+        when {
+            isArray ->
+                "Array<${kotlinTypeStringFor(componentType!!)}>"
+            isParameterized ->
+                "$parameterizedTypeDefinition<${actualTypeArguments.joinToString(transform = ::kotlinTypeStringFor)}>"
+            isWildcard ->
+                upperBound?.let(::kotlinTypeStringFor) ?: "Any"
+            else ->
+                toString().let { primitiveTypeStrings[it] ?: it }
```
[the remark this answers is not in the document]
eskatos: > Also, what happens if a package name has a keyword in it in Kotlin?
> Eg: com.when.something.AClass where when is a Kotlin keyword.

Very good point.
This PR doesn't touch this code, it simply moved it to another module.
Let's follow up on this in https://github.com/gradle/kotlin-dsl/issues/853

provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/ProjectSchemaAccessorsIntegrationTest.kt
eskatos: But integration coverage for a non-core plugin adding `SourceSet` conventions to account for the different classloading at play should indeed be added.

provider-plugins/src/main/kotlin/org/gradle/kotlin/dsl/provider/plugins/DefaultProjectSchemaProvider.kt
bamboo: :x: I'd duplicate it as `internal` to make the multiple usages of it in this module simpler and uniform with the rest of the codebase.

provider-plugins/src/main/kotlin/org/gradle/kotlin/dsl/provider/plugins/DefaultProjectSchemaProvider.kt
```diff
@@ -0,0 +1,120 @@
...
+
+private
+fun inferPublicTypeOfConvention(instance: Any) =
+    if (instance is HasPublicType) instance.publicType
+    else TypeOf.typeOf(instance::class.java.undecorated)
+
+
+private
+val Class<*>.undecorated
+    get() = takeIf { it.name.endsWith("_Decorated") }?.superclass ?: this
```
bamboo: :x: `name.endsWith("_Decorated")` seems very fragile. Maybe make the Gradle decorator add some marker interface or annotation to the types so they can be properly detected?

PR #855, opened 2018-05-02
Let the `kotlin-dsl` plugin add gradleKotlinDSl() to testImplementation

eskatos:
instead of `testRuntime` in order for the Kotlin DSL api to be available for tests implemented using `ProjectBuilder`, mocks etc...

### DIFF
--- a/plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslPlugin.kt
+++ b/plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslPlugin.kt
@@ -26,7 +26,7 @@ import org.gradle.kotlin.dsl.plugins.embedded.EmbeddedKotlinPlugin
  * The `kotlin-dsl` plugin.
  *
  * - Applies the `embedded-kotlin` plugin
- * - Adds the `gradleKotlinDsl()` dependency to the `compileOnly` and `testRuntimeOnly` configurations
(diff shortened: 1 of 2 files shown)


plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslPlugin.kt
```diff
@@ -38,7 +38,7 @@ open class KotlinDslPlugin : Plugin<Project> {
 
             applyEmbeddedKotlinPlugin()
             applyKotlinDslCompilerPlugins()
-            addGradleKotlinDslDependencyTo("compileOnly", "testRuntimeOnly")
+            addGradleKotlinDslDependencyTo("compileOnly", "testImplementation")
```
JLLeitschuh: What if a user isn't using the `java-library` plugin? Or is it automatically applied when this plugin is applied?

StefMa: I think the `embeddedKotlin` already applies the implementation somehow, right? 🤔

eskatos: @JLLeitschuh the `testImplementation` configuration is added by the [`java` plugin](https://docs.gradle.org/current/userguide/java_plugin.html), no need for the `java-library` plugin for it to exist. I think you mixed with the `api` configuration that is only available when using the `java-library` plugin.

Issue #492, opened 2017-09-12
TestKit. withPluginClasspath() issues when testing against `build.gradle.kts` scripts


mkobit:
TestKit failing to execute a build in the same way as importing the project locally.

### Expected Behavior

TestKit test should exhibit same behavior as standalone project and succeed.

### Current Behavior

Test fails unexpectedly, but works when running `publishToMavenLocal` and consuming the plugin from local `~/.m2/repository` in a test project succeeds (example project here https://github.com/mkobit/jenkins-pipeline-shared-library-example)
<details><summary>Build output</summary><p>

```
org.gradle.testkit.runner.UnexpectedBuildFailure: Unexpected build execution failure in /tmp/Kotlin DSL extension configuration$jenkins_pipeline_shared_libraries_gradle_plugin_main7624860855275780516 with arguments [-s, -i]

Output:
The client will now receive all logging from the daemon (pid: 605). The daemon log file: /tmp/.gradle-test-kit-circleci/test-kit-daemon/4.1/daemon-605.out.log
Starting 7th build in daemon [uptime: 30.778 secs, performance: 98%]
Using 36 worker leases.
Closing daemon's stdin at end of input.
The daemon will no longer process any standard input.
Creating new cache for fileHashes, path /tmp/Kotlin DSL extension configuration$jenkins_pipeline_shared_libraries_gradle_plugin_main7624860855275780516/.gradle/4.1/fileHashes/fileHashes.bin, access org.gradle.cache.internal.DefaultCacheAccess@29e6a1ff
Creating new cache for plugin-use-metadata, path /tmp/.gradle-test-kit-circleci/caches/4.1/plugin-resolution/plugin-use-metadata.bin, access org.gradle.cache.internal.DefaultCacheAccess@1c7aae0b
Creating new cache for client-status, path /tmp/.gradle-test-kit-circleci/caches/4.1/plugin-resolution/client-status.bin, access org.gradle.cache.internal.DefaultCacheAccess@1c7aae0b
Starting Build
Settings evaluated using settings file '/tmp/Kotlin DSL extension configuration$jenkins_pipeline_shared_libraries_gradle_plugin_main7624860855275780516/settings.gradle'.
Projects loaded. Root project using build file '/tmp/Kotlin DSL extension configuration$jenkins_pipeline_shared_libraries_gradle_plugin_main7624860855275780516/build.gradle.kts'.
[...]

PR #859, opened 2018-05-04
Compute script compilation classpath from target `ClassLoaderScope`

bamboo:
Instead of `scriptHandler.scriptClassPath` so plugin implementations coming from other loaders, as it is the case with plugins tested via TestKit's `withPluginClasspath`, are also added to the compilation classpath.

See #492


JLLeitschuh:
Thank you for finally fixing this!!!

PR #865, opened 2018-05-08
Optimise compilation classpath computation

bamboo:
- By caching the complete compilation `ClassPath` per `ClassLoaderScope`
- And by caching each individual `ClassLoader` `ClassPath`

The caches are part of a build scoped service and so all `ClassLoaderScope` and `ClassLoader` instances will be held until the end of the build.

### DIFF
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptClassPathProvider.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptClassPathProvider.kt
@@ -116,8 +116,8 @@ class KotlinScriptClassPathProvider(
         require(scope.isLocked) {
             "$scope must be locked before it can be used to compute a classpath!"
         }
-        val rootClassPath = cachedClassLoaderClassPath.of(scope.root.exportClassLoader)
         val fullClassPath = cachedClassLoaderClassPath.of(scope.exportClassLoader)
+        val rootClassPath = cachedClassLoaderClassPath.of(scope.root.exportClassLoader)
         return fullClassPath - rootClassPath
     }
 
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptClassPathProvider.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptClassPathProvider.kt
@@ -25,7 +25,7 @@ import org.gradle.api.internal.ClassPathRegistry
 import org.gradle.api.internal.artifacts.dsl.dependencies.DependencyFactory
 import org.gradle.api.internal.initialization.ClassLoaderScope
 
-import org.gradle.internal.classloader.ClasspathUtil.getClasspath
+import org.gradle.internal.classloader.ClassLoaderVisitor
 import org.gradle.internal.classpath.ClassPath
 import org.gradle.internal.classpath.DefaultClassPath
 
@@ -38,6 +38,10 @@ import org.gradle.kotlin.dsl.support.serviceOf
 import org.gradle.util.GFileUtils.moveFile
 
 import java.io.File
+
+import java.net.URI
+import java.net.URISyntaxException
+import java.net.URL
 
 
 fun gradleKotlinDslOf(project: Project): List<File> =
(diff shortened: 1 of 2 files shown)


bamboo:
Merging optimistically to get the benchmarks running on gradle/gradle.

bamboo:
With these changes the Gradle build performance test goes [from](https://builds.gradle.org/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=12682172&_focus=4554):
```
Speed Results for test project 'gradleBuildCurrent' with tasks help: Results were inconclusive with 93% confidence.
Difference: 31 ms slower (31 ms), 1.94%
  Current Gradle median: 1.632 s min: 1.569 s, max: 1.835 s, se: 68.48 ms}
```

[to](https://builds.gradle.org/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=12686710&_focus=3861):

```
Speed Results for test project 'gradleBuildCurrent' with tasks help: AWESOME! we're faster than 4.8-20180506235948+0000 with 99% confidence.
Difference: 80.5 ms faster (80.5 ms), -5.03%
  Current Gradle median: 1.518 s min: 1.401 s, max: 1.735 s, se: 77.2 ms}
``` 

And there seems to be no more obvious allocation hotspots.

bamboo:
Hey @oehme, do you see more opportunities for improvement here?

bamboo:
[Adhoc performance test scenario build](https://builds.gradle.org/viewLog.html?buildId=12686710&tab=buildResultsDiv&buildTypeId=Gradle_Util_Performance_AdHocPerformanceScenarioLinux).

oehme:
I'd do a global scoped cache with weak keys. There is no need to recompute this on every build and the cache entries will disappear as soon as that classloader goes unused.

bamboo:
> If we wanted to make this faster, I'd do a global scoped cache with weak keys.

Yeah, let's reconsider that when we do another optimisation pass.

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptClassPathProvider.kt
```diff
@@ -152,6 +162,12 @@ class KotlinScriptClassPathProvider(
     val gradleJars by lazy {
         classPathRegistry.getClassPath(gradleApiNotation.name).asFiles
     }
+
+    private
+    val cachedScopeCompilationClassPath = ConcurrentHashMap<ClassLoaderScope, ClassPath>()
+
+    private
+    val cachedClassLoaderClassPath = ClassLoaderClassPathCache()
```
JLLeitschuh: Am I reading this correctly? There is a double layer of caching going on here?

bamboo: Yes.

provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptClassPathProvider.kt
```diff
@@ -173,3 +189,53 @@ private
...
+    private
+    fun classPathOf(classLoader: ClassLoader): ClassPath {
+        val classPathFiles = mutableListOf<File>()
+
+        object : ClassLoaderVisitor() {
+            override fun visitClassPath(classPath: Array<URL>) {
+                classPath.forEach { url ->
+                    if (url.protocol == "file") {
+                        classPathFiles.add(fileFrom(url))
+                    }
```
JLLeitschuh: Are there any other kinds of protocols used by the class-loader besides "file"?

bamboo: > Are there any other kinds of protocols used by the class-loader besides "file"?

This is not related to this change which is a port from [ClasspathUtil.getClasspath](https://github.com/gradle/gradle/blob/96f28e5c54616f1af0fe38f943c302a0fd01f55d/subprojects/base-services/src/main/java/org/gradle/internal/classloader/ClasspathUtil.java#L57) with caching added thus the resulting `ClassPath` will be exactly the same as before.

JLLeitschuh: Understandable, I just wasn't aware that you could have the class loader work from anywhere but a file.

Issue #716, opened 2018-02-12
Members of specialized NamedDomainObjectContainer types should be available in their configuration block


eskatos:
For example, [TaskContainer#replace](https://docs.gradle.org/4.5/javadoc/org/gradle/api/tasks/TaskContainer.html#replace-java.lang.String-java.lang.Class-) is not available in the `tasks {}` block.

### Expected Behavior

```kotlin
tasks {
  replace("someTask").doLast {
  }
}
```

### Current Behavior

The above doesn't work as the receiver of the `tasks {}` block is `NamedDomainObjectContainer<Task>`, not `TaskContainer`.

Workaround is to move the statement out of the configuration block:

```kotlin
tasks.replace("someTask").doLast {
}
```

### Context
See #705 and #222

eskatos:
Superseded by #1000 
Closing

PR #867, opened 2018-05-10
Introduce `TaskContainer.getByName<T>`

bamboo:


eskatos:
❓can't we do this on `NamedDomainObjectContainer` and the corresponding scope instead?

bamboo:
We can't add it to `NamedDomainObjectContainer` because that would require two generic parameters. We might be able to add it to the scope but we have another issue to solve that holistically (#716).

eskatos:
What about the following?

```kotlin
inline fun <reified T : Any> NamedDomainObjectCollection<out Any>.getByName(name: String) =
    getByName(name).let {
        it as? T
            ?: throw IllegalStateException(
                "Element '$name' of type '${it::class.java.name}' cannot be cast to '${T::class.qualifiedName}'.")
    }

inline fun <reified T : Any> NamedDomainObjectCollection<out Any>.getByName(name: String, configure: T.() -> Unit) =
        getByName<T>(name).also(configure)
```

provider/src/main/kotlin/org/gradle/kotlin/dsl/TaskContainerExtensions.kt
```diff
@@ -25,3 +26,40 @@ import org.gradle.api.tasks.TaskContainer
...
+ *
+ * @param name task name
+ * @param configure configuration action to apply to the task before returning it
+ * @return task, never null
+ * @throws [UnknownDomainObjectException] When the given task is not found.
+ * @throws [IllegalStateException] When the given task cannot be cast to the expected type.
+ */
+@Suppress("extension_shadowed_by_member")
+inline fun <reified T : Any> TaskContainer.getByName(name: String, configure: T.() -> Unit) =
+    getByName<T>(name).also(configure)
```
JLLeitschuh: Does Kotlin still not support default parameters on inline functions?

bamboo: It does. I preferred not to mix the two concerns.

JLLeitschuh: Okay. Can you explain the mentality behind that a bit?
Cleaner code by making them two different methods?

provider/src/main/kotlin/org/gradle/kotlin/dsl/TaskContainerExtensions.kt
```diff
@@ -25,3 +26,40 @@ import org.gradle.api.tasks.TaskContainer
...
+ * @param name task name
+ * @return task, never null
+ * @throws [UnknownDomainObjectException] When the given task is not found.
+ * @throws [IllegalStateException] When the given task cannot be cast to the expected type.
+ */
+@Suppress("extension_shadowed_by_member")
+inline fun <reified T : Any> TaskContainer.getByName(name: String) =
+    getByName(name).let {
+        it as? T
+            ?: throw IllegalStateException(
```
JLLeitschuh: Wouldn't this be an `IllegalArgumentException`? Technically `T` is an argument correct?

provider/src/main/kotlin/org/gradle/kotlin/dsl/TaskContainerExtensions.kt
```diff
@@ -25,3 +26,40 @@ import org.gradle.api.tasks.TaskContainer
...
+ * If a task with the given [name] is not found, [UnknownDomainObjectException] is thrown.
+ * If the task is found but cannot be cast to the expected type [T], [IllegalStateException] is thrown.
+ *
+ * @param name task name
+ * @return task, never null
+ * @throws [UnknownDomainObjectException] When the given task is not found.
+ * @throws [IllegalStateException] When the given task cannot be cast to the expected type.
+ */
+@Suppress("extension_shadowed_by_member")
+inline fun <reified T : Any> TaskContainer.getByName(name: String) =
```
JLLeitschuh: Why isn't this on `TaskCollection`?

bamboo:
We decided to handle this differently.

bamboo:
Thanks for the review @JLLeitschuh. We're going to open a new PR soon with a different take on this.

JLLeitschuh:
Okay. Thanks.

Issue #833, opened 2018-04-20
Plugins should be defined in the plugins classloader


oehme:
The KotlinScriptBasePlugin and KotlinScriptRootPlugin (and maybe others I have missed) are currently part of Gradle's runtime classloader (because they are packed up in the `lib` directory), but should instead be part of the plugins classloader (packaged up in the `plugins` directory). This is important for the performance of plugin registry lookups. Basically to our plugin registry it looks like a plugin from an "unknown" location, triggering a slower lookup code path.

oehme:
The same is true for model builders btw, although they don't lead to performance problems, they still conceptually belong into the plugins directory, not the runtime.

eskatos:
Closing as delivered in #869 followed up by #871

PR #869, opened 2018-05-11
Let base plugins be defined in the Gradle Plugins classloader

eskatos:
by moving them to `:provider-plugins`, see #833


provider/src/main/kotlin/org/gradle/kotlin/dsl/accessors/DefaultAccessorsProvider.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import org.gradle.kotlin.dsl.provider.spi.AccessorsProvider
+import org.gradle.kotlin.dsl.provider.spi.ProjectSchema
+import org.gradle.kotlin.dsl.provider.spi.TypeAccessibility
+
+
+class DefaultAccessorsProvider : AccessorsProvider {
+
+    override fun forEachAccessorOf(accessibleSchema: ProjectSchema<TypeAccessibility>, action: (String) -> Unit) =
+        accessibleSchema.forEachAccessor(action)
```
JLLeitschuh: This seems like a somewhat excessive amount of indirection.

eskatos: Good catch! `:provider-plugins` could depend on `:provider` directly.
I'll open a simplification PR shortly.

PR #871, opened 2018-05-12
Simplify splitting :provider into :provider-plugins

eskatos:
by letting `:provider-plugins` depend directly on `:provider` and removing `:provider-spi`.

Follow up to #833 after #869


provider/src/main/kotlin/org/gradle/kotlin/dsl/accessors/DefaultAccessorsProvider.kt
```diff
@@ -1,28 +0,0 @@
...
-import org.gradle.kotlin.dsl.provider.spi.AccessorsProvider
-import org.gradle.kotlin.dsl.provider.spi.ProjectSchema
-import org.gradle.kotlin.dsl.provider.spi.TypeAccessibility
-
-
-class DefaultAccessorsProvider : AccessorsProvider {
-
-    override fun forEachAccessorOf(accessibleSchema: ProjectSchema<TypeAccessibility>, action: (String) -> Unit) =
-        accessibleSchema.forEachAccessor(action)
-}
```
JLLeitschuh: 👍

Issue #761, opened 2018-03-15
Top-level extensions of *ObjectContainer type should get the "invoke" style DSL


mkobit:
### Expected Behavior

Should be able to use them like other top level containers like `ConfigurationContainer` and also have parity to Groovy.

### Current Behavior

Have to do workaround demonstrated in https://github.com/gradle/kotlin-dsl/issues/459

### Context

Using container-types in plugins as top level types for better DSLs.

### Steps to Reproduce (for bugs)

Don't have full setup right now but here is a close-to example

1. Add a `buildSrc` plugin/`java-gradle-plugin` ceremony

    ```kotlin
    import org.gradle.api.Named
    import org.gradle.api.NamedDomainObjectContainer
    import org.gradle.api.Plugin
    import org.gradle.api.Project
    import org.gradle.api.reflect.TypeOf
    
    open class MkobitExample : Plugin<Project> {
      override fun apply(target: Project) {
        val myContainer = target.container(MyType::class.java)
        target.extensions.add("withoutType", myContainer)
        val type = object : TypeOf<NamedDomainObjectContainer<MyType>>() {}
        target.extensions.add(type,"withType", myContainer)
      }
    }
    
    data class MyType(private val myName: String) : Named {
      override fun getName(): String = myName
    }
    ```

2. Add in consumption to root `build.gradle.kts`

    ```kotlin
    plugins {
      id("mktest")
    }
    withoutType {
      "hello" {
    
      }
    }
    
    withType {
      "hi" {
    
      }
    }
    ```

```
e: /home/mkobit/dotfiles/build.gradle.kts:148:7: Expression '"hello"' of type 'String' cannot be invoked as a function. The function 'invoke()' is not found
e: /home/mkobit/dotfiles/build.gradle.kts:154:7: Expression '"hi"' of type 'String' cannot be invoked as a function. The function 'invoke()' is not found


FAILURE: Build failed with an exception.

* Where:
Build file '/home/mkobit/Workspace/code_lab/example/build.gradle.kts' line: 148

* What went wrong:
Script compilation errors:

  Line 148:       "hello" {
                  ^ Expression '"hello"' of type 'String' cannot be invoked as a function. The function 'invoke()' is not found

  Line 154:       "hi" {
                  ^ Expression '"hi"' of type 'String' cannot be invoked as a function. The function 'invoke()' is not found
```

### Your Environment

Gradle 4.6 

 * Build scan URL:

eskatos:
[...]

PR #618, opened 2017-11-28
Various reified extensions on NamedDomainObjectContainerExtensions

JLLeitschuh:
Closes #605

### Contributor Checklist
- [x] Base the PR against the `develop` branch
- [x] [Sign Gradle CLA](http://gradle.org/contributor-license-agreement/)
- [ ] Provide integration tests to verify changes from a user perspective
- [x] Provide unit tests to verify logic
- [ ] Ensure that tests pass locally: `./gradlew check --parallel`

Issue #804, opened 2018-04-03
Make `PolymorphicDomainObjectContainer` methods and extensions available in collection configuration blocks


bamboo:
```kotlin
tasks {
    create<MyTask>("my") // should work
}
```

Relates to gradle/gradle#4570

Issue #605, opened 2017-11-22
NamedDomainObjectContainerExtensions should have reified versions of API's


JLLeitschuh:
### Expected Behavior

You should be able to do things like:
```kotlin
val test by tasks.getting<Test> { useTestNG() }
```

Currently, the API only supports using the `NamedDomainObjectContainerExtensions` with API's requiring the `Klass` type.
So currently you have to do this:
```kotlin
val test by tasks.getting(Test::class) { useTestNG() }
```

### Context
The requirement to pass the `Klass` type is very un-kotliny. The reified type reads better and also allows for this sort of syntax:

```kotlin
val test : Test by tasks.getting { useTestNG() }
```

### Your Environment
Current API: https://github.com/gradle/kotlin-dsl/blob/6a3c51a65e8423aa33432b816996360cd32e7906/provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectContainerExtensions.kt

bamboo:
The PR doesn't actually allow `val test by tasks.getting<Test> { useTestNG() }`, right? It makes sense to create a follow-up issue.

JLLeitschuh:
I mean, for that I always do `val task = task<Test>("test") { useTestNG() }`. I suppose that we should support for the above API yes.

Is the `getting` on `tasks` explicitly declared somewhere else?

bamboo:
It's declared as an extension to `PolymorphicDomainObjectContainer` in https://github.com/gradle/kotlin-dsl/blob/6b27df709ccd65c52ffac86c46c1fe5d67620d94/provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectContainerExtensions.kt#L191

JLLeitschuh:
Yea, unfortunately there's no good way that I could think up to come up with a reified version of that overload that the compiler would be happy with.

If you have any clever thoughts in that area I'm totally looking for solutions.

Every generic combo I tried all caused some other test to not compile.

mkobit:
I also added a comment at https://github.com/gradle/kotlin-dsl/issues/818#issuecomment-395458741 with some of the current inconsistencies and UX when trying to use the existing `getting`/`creating` APIs if you don't explicitly provide the `KClass<>` as an argument.

PR #873, opened 2018-05-12
Refine [Named|Polymorphic]DomainObject[Collection|Container] extensions

eskatos:
by making their whole API available with the invoke syntax
and adding reified getByName<T>(name) {} to NamedDomainObjectCollection
and adding reified maybeCreate<T>(name) to PolymorphicDomainObjectContainer
adding more coverage along the way

See #804 and #867

### DIFF
--- a/provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectContainerExtensions.kt
+++ b/provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectContainerExtensions.kt
@@ -16,6 +16,7 @@
 
 package org.gradle.kotlin.dsl
 
+import org.gradle.api.Action
 import org.gradle.api.NamedDomainObjectContainer
 import org.gradle.api.PolymorphicDomainObjectContainer
 
@@ -46,7 +47,19 @@ inline operator fun <T : Any, C : NamedDomainObjectContainer<T>> C.invoke(
  */
 class NamedDomainObjectContainerScope<T : Any>(
     private val container: NamedDomainObjectContainer<T>
-) : NamedDomainObjectContainer<T> by container {
+) : NamedDomainObjectContainer<T> by container, PolymorphicDomainObjectContainer<T> {
+
+    override fun <U : T> create(name: String, type: Class<U>): U =
+        polymorphicDomainObjectContainer().create(name, type)
+
+    override fun <U : T> create(name: String, type: Class<U>, configuration: Action<in U>): U =
+        polymorphicDomainObjectContainer().create(name, type, configuration)
+
+    override fun <U : T> maybeCreate(name: String, type: Class<U>): U =
+        polymorphicDomainObjectContainer().maybeCreate(name, type)
+
+    override fun <U : T> containerWithType(type: Class<U>): NamedDomainObjectContainer<U> =
+        polymorphicDomainObjectContainer().containerWithType(type)
 
     /**
      * @see [NamedDomainObjectContainer.maybeCreate]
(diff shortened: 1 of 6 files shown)


mkobit:
Nice! I think this also will fix https://github.com/gradle/kotlin-dsl/issues/761

Another thing related to this is better `invoke` style DSL that I mentioned at https://github.com/gradle/kotlin-dsl/issues/459#issuecomment-373759896

provider/src/main/kotlin/org/gradle/kotlin/dsl/support/Exceptions.kt
```diff
@@ -22,7 +22,7 @@ import kotlin.reflect.KClass
 
 
 fun illegalElementType(container: NamedDomainObjectCollection<*>, name: String, expectedType: KClass<*>, actualType: KClass<*>) =
```
JLLeitschuh: Why not make an exception called `IllegalElementType` which extends `IllegalArgumentException`? Wouldn't that be more descriptive to the user?

JLLeitschuh:
Does this supersede my PR #618?

provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectContainerExtensions.kt
```diff
@@ -46,7 +44,19 @@ inline operator fun <T : Any, C : NamedDomainObjectContainer<T>> C.invoke(
...
+        polymorphicDomainObjectContainer().create(name, type)
+
+    override fun <U : T> create(name: String, type: Class<U>, configuration: Action<in U>): U =
+        polymorphicDomainObjectContainer().create(name, type, configuration)
+
+    override fun <U : T> maybeCreate(name: String, type: Class<U>): U =
+        polymorphicDomainObjectContainer().maybeCreate(name, type)
+
+    override fun <U : T> containerWithType(type: Class<U>): NamedDomainObjectContainer<U> =
+        polymorphicDomainObjectContainer().containerWithType(type)
```
JLLeitschuh: Don't you want this to return a `NamedDomainObjectContainerScope`?

provider/src/test/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectContainerExtensionsTest.kt
```diff
@@ -98,20 +166,36 @@ class NamedDomainObjectContainerExtensionsTest {
...
         }
 
         tasks {
+            create<Delete>("clean") {
+                delete("some")
+            }
+            getByName<Delete>("clean") {
+                delete("stuff")
+            }
             "clean"(type = Delete::class) {
```
JLLeitschuh: Did not realize that this syntax existed. Interesting.

provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectContainerExtensions.kt
eskatos: A simple `NamedDomainObjectContainer` is enough, the scope for the returned container will be provided by https://github.com/gradle/kotlin-dsl/pull/873/files/ade1cc6b0cf0287ea074791a86e599df53bc2a4a#diff-3527cb19f2f40291861f41b1ffe878bfR33

```kotlin
container.containerWithType(...) {
    // scoped
}
// or
val other = container.containerWithType(...)
other {
    // scoped
}
```

eskatos:
> Does this supersede my PR #618?

It doesn't. #618 is about #605 which isn't actionable until #605 is fixed.
I'll close #618 for now as the solution shouldn't introduce any asymmetry in the DSL and will be a bit different.

PR #882, opened 2018-05-16
Provide correct classpath for script plugins with buildscript blocks

bamboo:


provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/BuildServices.kt
```diff
@@ -66,6 +67,24 @@ object BuildServices {
     fun createClassPathModeExceptionCollector() =
         ClassPathModeExceptionCollector()
 
+    @Suppress("unused")
```
JLLeitschuh: Can you explain why this is here but unused?

PR #887, opened 2018-05-20
Optimise script execution via partial evaluation techniques

bamboo:
# A new Kotlin DSL interpreter

An optimised interpreter for the Kotlin DSL based on the idea of [partial evaluation](https://en.wikipedia.org/wiki/Partial_evaluation).

Instead of interpreting a given Kotlin DSL script directly, the interpreter emits a specialized program that captures the optimal execution procedure for the particular combination of script structure (does it contain a `buildscript` block? a `plugins` block? a script body? etc), target object and script kind (top-level or plugin).

The specialized program is then cached via a cheap cache key based on the original, unprocessed contents of the script, target context and parent `ClassLoader`.

Because each program is specialized to a given script structure, a lot of work is avoided. For example, a top-level script containing a `plugins` block but no body can be compiled down to a specialized program that instantiates the precompiled `plugins` block class directly - without reflection - and does nothing else. The same strategy can be used for a **script plugin** (a non top-level script) with a body but no `buildscript` block since the classpath is completely determined at the time the specialized program is emitted.

## Must
- [x] evaluator for empty settings
- [x] evaluator for stage2 settings
- [x] evaluator for stage1 settings
- [x] serialize HashCode in the bytecode
- [x] keep preprocessed file in disk
- [x] evaluator for staged settings script
- [x] make sure there are no dependencies from execution into provider
- [x] replace main evaluator for settings scripts
  - [x] fix location aware handling for settings script
- [x] stage 2 only non-top level scripts (script plugins) can be executed as precompiled scripts
- [x] introduce ProgramType.{TopLevel,ScriptPlugin}
- [x] introduce ProgramTarget.{Project,Settings}
- [x] replace main evaluator for project script plugins
- [x] measure
  - [x] create bamboo/release/partial-evaluator-snapshot branch
  - [x] publish 0.18.x snapshot from that branch
  - [x] update kotlin-dsl-snapshto branch
  - [x] run adhoc target on that branch
- [x] evaluator for project scripts with only stage1 plugins
- [x] evaluator for top-level project scripts with stage1 buildscript block (accessors classpath)
- [x] pass `accessorsClassPath` through
- [x] evaluator for project scripts with both a buildscript and a plugins block as a single script
- [x] evaluator for top-level project script with stage2 (accessor classpath)
- [x] replace main evaluator for project scripts
- [x] measure
- [ ] Introduce `prepareScriptHandler` (for `setupEmbeddedKotlinRepository`)
- [ ] measure
- [x] Refactor `ResidualProgramCompiler` into two steps with a reified `SpecializedProgram` representation in between for simpler code and tests
- [x] preserve line numbers on `plugins` blocks
- [x] replace main evaluator for init scripts
- [ ] measure
- [ ] precompute stage 2 `childClassLoaderScopeId`
- [ ] introduce SpecializedProgramCache with an optimised cache key representing the tuple (programTarget, programKind, sourceHash, accessorsClassPathHash, weakParentClassLoader)
  - [ ] synthesize cache key at compilation time? a single `Long` could replace `scriptTemplateId`
  - [ ] consider introducing separate SpecializedProgramCache for stage 2
- [ ] measure
- [x] single Lexer pass over the whole script collecting comments and top-level blocks by name
- [ ] measure
- [x] use a `ProgramHost` implementation that ignores errors for classpath mode
- [ ] reconsider whether a top-level `Program.Script` for `Settings` should reduce to `Dynamic` 

## Open Questions

- Does a ProgramKind.ScriptPlugin without a Stage2 need to `closeTargetScopeOf` ?

## Could

- [x] simplify KotlinScriptClassloadingCache::loadScriptClass
  - [x] remove ClassLoaderScope references
  - [x] replace ClassPath by HashCode and precompute HashCode


bamboo:
Hey @eskatos, this is not ready for a full review yet but I'm pushing it in case you're curious.

By the way, I could definitely use some help reviewing the ignored tests in`ScriptCachingIntegrationTest` :)

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Program.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    data class Plugins(val fragment: ProgramSourceFragment) : Stage1()
+
+    data class Stage1Sequence(val buildscript: Buildscript, val plugins: Plugins) : Stage1()
+
+    data class Script(val source: ProgramSource) : Program()
+
+    data class Staged(val stage1: Stage1, val stage2: Script) : Program()
+
+    abstract class Stage1 : Program()
```
JLLeitschuh: Can you elaborate on what a `Stage1` is?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/ProgramSource.kt
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Normalised program text guaranteed to use `\n` as the only line terminator character.
+ */
+data class ProgramText private constructor(val text: String) {
+
+    companion object {
+
+        fun from(string: String) =
+            ProgramText(string.replace("\r\n", "\n"))
```
JLLeitschuh: You only care about Windows an Unix systems here? You don't care about the other possible line terminators correct?
https://en.wikipedia.org/wiki/Newline#Representations_in_different_character_encoding_specifications

JLLeitschuh:
So, great idea. Just a thought though, sometimes my `build.gradle.kts` files are as simple as just doing:
```
description =
    "Sub-projects are dedicated to exposing an API for performing fit calculations and operations."
```

So as soon as I add a `description` to my project's `build.gradle.kts` I loose access to this optimization correct?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/ProgramSource.kt
bamboo: Correct.

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/execution/InterpreterTest.kt
```diff
@@ -0,0 +1,189 @@
...
+                verify(host).cache(
+                    stage2TemplateId,
+                    sourceHash,
+                    targetScopeLocalClassLoader,
+                    classLoaders[1].loadClass("Program"))
+            }
+        } finally {
+            classLoaders.forEach {
+                it.close()
+            }
```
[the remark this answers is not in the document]
bamboo: Thanks, that's ok.

bamboo:
> So as soon as I add a description to my project's build.gradle.kts I loose access to this optimization correct?

Only to the `Empty` program optimisation, after all it's no longer empty, but script execution would still take advantage of the _specialized program_ optimisation.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/PartialEvaluator.kt
```diff
@@ -20,12 +20,24 @@ import org.jetbrains.kotlin.lexer.KotlinLexer
...
+enum class ProgramKind {
+    TopLevel,
+    ScriptPlugin
+}
+
+
+enum class ProgramTarget {
+    Project,
+    Settings
+}
```
JLLeitschuh: Doesn't this limit scripts that are applied to a different target type?
```
apply "foo.gradle.kts" to something
```

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Program.kt
JLLeitschuh: Thanks!!!

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/PartialEvaluator.kt
bamboo: For now only `Project` and `Settings` are supported.

In general, if you apply the same script to different targets, the interpreter will specialise the program once for each target.

JLLeitschuh: I understand that only `Project` and `Settings` is currently supported, I just want to make sure that this code isn't somehow limiting future support for other targets.

Also, how does this handle one script that is both targeted at being a `Settings` and a `Project` target. This is a setup I have in my repo today and I want to make sure that this use case isn't broken by this.

bamboo: > I just want to make sure that this code isn't somehow limiting future support for other targets.

Rest assured it isn't :)

> This is a setup I have in my repo today and I want to make sure that this use case isn't broken by this.

It should continue to work as before.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptEvaluator.kt
```diff
@@ -207,4 +128,108 @@ class StandardKotlinScriptEvaluator(
...
+        override fun cache(
+            templateId: String,
+            sourceHash: HashCode,
+            parentClassLoader: ClassLoader,
+            specializedProgram: Class<*>
+        ) {
+            classloadingCache.put(
+                cacheKeyFor(templateId, sourceHash, parentClassLoader),
+                LoadedScriptClass(
+                    CompiledScript(File(""), specializedProgram.name, Unit),
```
JLLeitschuh: Why an empty file?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptEvaluator.kt
```diff
@@ -207,4 +128,108 @@ class StandardKotlinScriptEvaluator(
...
+            childScopeId: String,
+            location: File,
+            className: String,
+            accessorsClassPath: ClassPath?
+        ): Class<*> =
+            classLoaderScope
+                .createChild(childScopeId)
+                .local(DefaultClassPath.of(location))
+                .apply { accessorsClassPath?.let(::local) }
+                .lock()
```
JLLeitschuh: Pre-mature locking of a `ClassLoaderScope` has caused a few issues. Just want to make sure those potential issues have been evaluated here.

#4823
#311

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/CharSequenceExtensions.kt
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package org.gradle.kotlin.dsl.provider
+package org.gradle.kotlin.dsl.execution
```
JLLeitschuh: API breaking change?

Barteks2x:
Will it make running a script for the first time (ie. the script has just been modified a while ago) faster or slower? As I understand it, it's intended to make running an already compiled script faster. Most of the time when I'm running gradle task is when working in the buildscript, which means frequently modifying it. After that, about the only usages of buildscript for me are importing everything into IDE and running it on CI server.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/ProgramParser.kt
```diff
@@ -0,0 +1,118 @@
...
+        val comments = mutableListOf<IntRange>()
+        start(script)
+        while (tokenType != null) {
+            if (tokenType in KtTokens.COMMENTS) {
+                comments.add(tokenStart..(tokenEnd - 1))
+            }
+            advance()
+        }
+        comments
+    }
```
JLLeitschuh: Is the `IntRange` the lines that the block comments are on? Does this handle non-block comments?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/PartialEvaluator.kt
```diff
@@ -30,108 +33,143 @@ enum class ProgramKind {
 
 enum class ProgramTarget {
     Project,
-    Settings
+    Settings,
+    Gradle
```
JLLeitschuh: What is `Gradle` in this? Also, what about `InitScript` support?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/ProgramParser.kt
bamboo: Yes, see `ProgramParserTest`.

JLLeitschuh:
High level question, is there any part of this that would be more useful if contributed directly to the Kotlin compiler project? Do any of these optimizations have a wider reach than just for the Gradle use case?
Does it make sense to try to work some of those into their compiler instead?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/PartialEvaluator.kt
bamboo: From the point of the view of the Kotlin DSL, an init script is script that targets the `Gradle` singleton.

bamboo:
Hey @Barteks2x,

> Will it make running a script for the first time (ie. the script has just been modified a while ago) faster or slower?

Depending on the structure of the script it will make them faster or the same as before.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptEvaluator.kt
bamboo: Thanks. That's the same behaviour as before.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Interpreter.kt
```diff
@@ -477,5 +509,19 @@ fun maybeUnwrapInvocationTargetException(e: Throwable) =
...
+inline fun withContextClassLoader(classLoader: ClassLoader, block: () -> Unit) {
+    val currentThread = Thread.currentThread()
+    val previous = currentThread.contextClassLoader
+    try {
+        currentThread.contextClassLoader = classLoader
+        block()
+    } finally {
+        currentThread.contextClassLoader = previous
+    }
+}
```
JLLeitschuh: Can you explain what's going on here a bit more? Are you just able to swap out the class-loader on a thread like this?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/ResidualProgramCompiler.kt
```diff
@@ -467,6 +467,10 @@ class ResidualProgramCompiler(
         return compileScript(scriptFile, originalScriptPath, scriptDefinition)
     }
 
+    init {
+        org.jetbrains.kotlin.cli.common.environment.setIdeaIoUseFallback()
+    }
```
JLLeitschuh: It seems kind of weird that this `init` block is hidden away in the middle of the file instead of at the top.

Also, it seems weird that you are using a constructor to change what seems like some static state about the system.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptEvaluator.kt
```diff
@@ -255,3 +245,13 @@ private
...
     transitiveClosureOf("stdlib-jdk8", "reflect")
 }
+
+
+private
+inline fun ClassPathModeExceptionCollector.ignoringErrors(action: () -> Unit) =
+    try {
+        action()
+    } catch (e: Exception) {
+        e.printStackTrace()
```
JLLeitschuh: Use a logger?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptPluginFactory.kt
```diff
@@ -24,11 +24,15 @@ import org.gradle.configuration.ScriptPluginFactory
...
 
+import org.gradle.kotlin.dsl.execution.EvalOption
+import org.gradle.kotlin.dsl.execution.defaultEvalOptions
+
 import java.util.*
 
 import javax.inject.Inject
 
 
+@Suppress("unused") // This name of this class is hardcoded in Gradle
```
JLLeitschuh: Thank you for the comment here!

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
```diff
@@ -146,12 +146,19 @@ fun compileTo(
...
+            return compileKotlin(environment)
         }
     }
 }
 
 
+private
+fun compileKotlin(environment: KotlinCoreEnvironment): Boolean {
+    org.jetbrains.kotlin.cli.common.environment.setIdeaIoUseFallback()
+    return compileBunchOfSources(environment)
```
JLLeitschuh: Is there a way to statically disable this after this is called? Do you want to reset the static state back?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Lexer.kt
```diff
@@ -17,118 +17,151 @@
...
+                                            )
+                                        )
+                                        reset()
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
```
JLLeitschuh: Consider adding some documentation explaining what you are looking for here.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptEvaluator.kt
```diff
@@ -0,0 +1,254 @@
...
+        options: EvalOptions
+    ) {
+
+        // TODO:partial-evaluator Optimise away this call whenever possible via partial evaluation
+        setupEmbeddedKotlinForBuildscript(scriptHandler)
+
+        val sourceHash = scriptSourceHasher.hash(scriptSource)
+
+        if (EvalOption.IgnoreErrors in options)
+            classPathModeExceptionCollector.ignoringErrors {
```
eskatos: :x: Extract condition and error collection in its own function in order not to repeat the `eval()` call

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/KotlinSettingsScriptIntegrationTest.kt
```diff
@@ -39,6 +40,7 @@ class KotlinSettingsScriptIntegrationTest : AbstractIntegrationTest() {
             containsString("*42*"))
     }
 
+    @Ignore("partial-evaluator-wip")
```
eskatos: :x: regression must be fixed before merging

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/integration/ScriptCachingIntegrationTest.kt
```diff
@@ -38,6 +39,7 @@ class ScriptCachingIntegrationTest : AbstractIntegrationTest() {
         val isolatedTestKitDir = IsolatedTestKitDir()
     }
 
+    @Ignore("partial-evaluator-wip")
```
eskatos: I'll have a look at fixing those tests, a good way to get to know the partial evaluator.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Interpreter.kt
```diff
@@ -0,0 +1,535 @@
...
+    }
+
+
+private
+fun maybeUnwrapInvocationTargetException(e: Throwable) =
+    if (e is InvocationTargetException) e.targetException
+    else e
+
+
+// TODO:partial-evaluator restore contextClassLoader behaviour
```
eskatos: :x: must be fixed before merging

bamboo: Done.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/EvalOption.kt
```diff
@@ -0,0 +1,33 @@
...
+    SkipBody
+}
+
+
+typealias EvalOptions = EnumSet<EvalOption>
+
+
+internal
+val defaultEvalOptions: EvalOptions =
+    EnumSet.noneOf(EvalOption::class.java)
```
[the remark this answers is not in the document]
bamboo: To keep the public surface to a minimum.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptEvaluator.kt
```diff
@@ -207,4 +128,108 @@ class StandardKotlinScriptEvaluator(
...
             embeddedKotlinProvider,
             classPathModeExceptionCollector)
+
+    private
+    val interpreter by lazy {
+        Interpreter(InterpreterHost())
+    }
+
+    private
+    inner class InterpreterHost : Interpreter.Host {
```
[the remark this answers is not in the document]
bamboo: The `Host` represents the set of services required by the `Interpreter` from its environment.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/CharSequenceExtensions.kt
bamboo: These functions are not part of the API.

build.gradle.kts
```diff
@@ -112,6 +114,7 @@ val benchmark by task<integration.Benchmark> {
 
 // -- IntelliJ IDEA configuration --------------------------------------
 idea {
+    (this as IdeaModel)
```
JLLeitschuh: Why is this now necessary?

eskatos: This is caused by a change in `gradle/gradle`.
That's a temporary measure until https://github.com/gradle/gradle/pull/5644 is merged.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptEvaluator.kt
eskatos: This codepath only happen through the IntelliJ script dependencies resolver, which will log those collected exceptions. No need to print the trace here.

:x: remove that line

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptEvaluator.kt
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        override fun applyBasePluginsTo(project: Project) {
+            kotlinScriptBasePluginsApplicator
+                .apply(project)
+        }
+
+        override fun closeTargetScopeOf(scriptHost: KotlinScriptHost<*>) {
+            val targetScope = scriptHost.targetScope
+//                    targetScope.export(classPathProvider.gradleApiExtensions)
```
eskatos: :x: remove commented code

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/PluginRequestsHandler.kt
```diff
@@ -31,7 +31,7 @@ import javax.inject.Inject
 
 internal
 class PluginRequestsHandler @Inject constructor(
-    private val pluginRequestApplicator: PluginRequestApplicator,
+    val pluginRequestApplicator: PluginRequestApplicator,
```
eskatos: :x: make it `internal`

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/architecture/PackageDependencyRulesTest.kt
```diff
@@ -0,0 +1,58 @@
...
+package org.gradle.kotlin.dsl.architecture
+
+import com.tngtech.archunit.core.importer.ClassFileImporter
+import com.tngtech.archunit.lang.syntax.ArchRuleDefinition.noClasses
+
+
+import org.junit.Test
+
+
+class PackageDependencyRulesTest {
```
eskatos: 👍

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Lexer.kt
eskatos: Explaining comment is already present: https://github.com/gradle/kotlin-dsl/pull/887/files#diff-ad9df6da18fa73c22b63aa9b5a7c0921R39

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/ResidualProgramCompiler.kt
```diff
@@ -0,0 +1,754 @@
...
+            outputDir,
+            scriptFile,
+            scriptDefinition,
+            classPath.asFiles,
+            messageCollectorFor(logger) { path ->
+                if (path == scriptFile.path) originalPath
+                else path
+            })
+
+    /* TODO:partial-evaluator put progress logging back
```
eskatos: :x: must be fixed before merging

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Interpreter.kt
```diff
@@ -0,0 +1,559 @@
...
+ * `plugins` block class directly - without reflection - and does nothing else. The
+ * same strategy can be used for a **script plugin** (a non top-level script) with a
+ * body but no `buildscript` block since the classpath is completely determined at
+ * the time the specialized program is emitted.
+ *
+ * @see Program
+ * @see PartialEvaluator
+ * @see ResidualProgram
+ * @see ResidualProgramCompiler
+ */
```
JLLeitschuh: 👍

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Interpreter.kt
```diff
@@ -0,0 +1,559 @@
...
+        programHost.eval(specializedProgram, scriptHost)
+    }
+
+    private
+    fun programTargetFor(target: Any): ProgramTarget =
+        when (target) {
+            is Settings -> ProgramTarget.Settings
+            is Project -> ProgramTarget.Project
+            is Gradle -> ProgramTarget.Gradle
+            else -> throw IllegalArgumentException("Unsupported target: $target")
```
JLLeitschuh: :o: Do you want to print the `$target` or the `${target::class}`?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Interpreter.kt
```diff
@@ -0,0 +1,559 @@
...
+        val cachedDir =
+            host.cachedDirFor(templateId, sourceHash, parentClassLoader) { cachedDir ->
+
+                logCompilationOf(templateId, scriptSource)
+
+                val outputDir =
+                    stage1SubDirOf(cachedDir).apply { mkdir() }
+
+                val sourceText =
+                    scriptSource.resource!!.text
```
JLLeitschuh: :o: Why is this known not to be null? Might be good to mention in a comment?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Interpreter.kt
```diff
@@ -0,0 +1,559 @@
...
+    fun eval(
+        target: Any,
+        scriptSource: ScriptSource,
+        sourceHash: HashCode,
+        scriptHandler: ScriptHandler,
+        targetScope: ClassLoaderScope,
+        baseScope: ClassLoaderScope,
+        topLevelScript: Boolean,
+        options: EvalOptions = defaultEvalOptions
+    ) {
```
JLLeitschuh: When you've reached 8 arguments it might be time to consider refactoring your args into a class/builder.
Something like `EvalArgs`.
I notice that the `evaluate` above has pretty much the same arguments as the `eval` here. If you end up needing more arguments, this is just going to mean two places to update.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Interpreter.kt
```diff
@@ -0,0 +1,559 @@
...
+            if (EvalOption.SkipBody in options) return
+            super.evaluateSecondStageOf(program, scriptHost, scriptTemplateId, sourceHash)
+        }
+
+        override fun handleScriptException(
+            exception: Throwable,
+            scriptClass: Class<*>,
+            scriptHost: KotlinScriptHost<*>
+        ) {
+            // TODO:partial-evaluator let the host collect this exception
```
JLLeitschuh: :o: Resolve before merging?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Interpreter.kt
```diff
@@ -0,0 +1,559 @@
...
+            return exception
+        }
+
+        exception.stackTrace.find(::scriptStackTraceElement)?.run {
+            return LocationAwareException(original, scriptSource, lineNumber.takeIf { it >= 0 })
+        }
+
+        val cause = exception.cause ?: return null
+        return inferLocationFrom(cause)
+    }
```
JLLeitschuh: So I understand _what_ this is doing, but I don't understand _why_ you are doing it. Can you add a bit of a comment explaining?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Lexer.kt
JLLeitschuh: You are totally correct, my mistake.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Lexer.kt
```diff
@@ -0,0 +1,167 @@
...
+import org.jetbrains.kotlin.lexer.KotlinLexer
+import org.jetbrains.kotlin.lexer.KtTokens.COMMENTS
+import org.jetbrains.kotlin.lexer.KtTokens.IDENTIFIER
+import org.jetbrains.kotlin.lexer.KtTokens.LBRACE
+import org.jetbrains.kotlin.lexer.KtTokens.RBRACE
+import org.jetbrains.kotlin.lexer.KtTokens.WHITE_SPACE
+
+
+internal
+class UnexpectedBlock(val identifier: String, val location: IntRange) : RuntimeException("Unexpected block found.")
```
JLLeitschuh: Do you want to include the `identifier` and the `location` in the exception message?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Program.kt
```diff
@@ -0,0 +1,76 @@
...
+ * - a script with neither a buildscript nor a plugins block
+ * - a script preceded by a buildscript or plugins block or both
+ *
+ * The evaluation of a Kotlin DSL program happens, in the general case, in
+ * [two stages](https://en.wikipedia.org/wiki/Multi-stage_programming):
+ * - in stage 1, the [Buildscript] and [Plugins] programs are executed and their execution
+ *   is assumed to affect the classpath available to the stage 2 program;
+ * - in stage 2, the remaining [Script] must be evaluated against the dynamically resolved classpath and,
+ *   for that reason, [stage 2 programs][Script] can only be specialized after stage 1 executes at least once;
+ */
```
JLLeitschuh: 👍 This makes this PR Make quite a bit more sense. Thanks!

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/ProgramParser.kt
```diff
@@ -0,0 +1,127 @@
...
+                Program.Staged(s1, s2)
+            } ?: s1
+        }
+
+        return stage2
+            ?: Program.Empty
+    }
+
+    private
+    fun checkForSingleBlocksOf(
```
JLLeitschuh: To help out with documentation, it might be helpful to document that this method will throw a `UnexpectedBlock` if this check fails.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/ProgramParser.kt
```diff
@@ -0,0 +1,127 @@
...
+
+        val plugins =
+            pluginsFragment?.takeIf { it.isNotBlank() }?.let(Program::Plugins)
+
+        val stage1 =
+            buildscript?.let { bs ->
+                plugins?.let { ps ->
+                    Program.Stage1Sequence(bs, ps)
+                } ?: bs
+            } ?: plugins
```
JLLeitschuh: Took me a few seconds to figure out what was going on here. But the end result is quite clever.
It takes a bit of time to start reading `?.let {` as an if statement.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/ProgramParser.kt
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    private
+    fun checkForSingleBlocksOf(
+        topLevelBlockIds: Array<String>,
+        topLevelBlocks: List<TopLevelBlock>
+    ) {
+        topLevelBlockIds.forEach { id ->
+            topLevelBlocks
+                .filter { it.identifier === id }
```
JLLeitschuh: Why reference equality?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/Lexer.kt
JLLeitschuh: Can you rename this to `UnexpectedBlockException` for clarity?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/ProgramParser.kt
```diff
@@ -0,0 +1,127 @@
...
+fun handleUnexpectedBlock(unexpectedBlock: UnexpectedBlock, script: String, scriptPath: String): Nothing {
+    val (line, column) = script.lineAndColumnFromRange(unexpectedBlock.location)
+    val message = compilerMessageFor(scriptPath, line, column, unexpectedBlockMessage(unexpectedBlock))
+    throw IllegalStateException(message, unexpectedBlock)
+}
+
+
+private
+fun unexpectedBlockMessage(block: UnexpectedBlock) =
+    "Unexpected `${block.identifier}` block found. Only one `${block.identifier}` block is allowed per script."
```
JLLeitschuh: Why not put this as the message in/on the `UnexpectedBlock`

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/ResidualProgramCompiler.kt
```diff
@@ -0,0 +1,754 @@
...
+                "Ljava/lang/String;Ljava/lang/String;" +
+                "Lorg/gradle/kotlin/dsl/support/KotlinScriptHost;" +
+                "Ljava/lang/String;Lorg/gradle/internal/hash/HashCode;" +
+                "Lorg/gradle/kotlin/dsl/execution/ProgramKind;" +
+                "Lorg/gradle/kotlin/dsl/execution/ProgramTarget;" +
+                ")Ljava/lang/Class;")
+    }
+
+    private
+    fun ClassVisitor.overrideExecute(methodBody: MethodVisitor.() -> Unit) {
```
JLLeitschuh: So I'm seeing this called a bunch of places, but I have no idea what it's doing or why. Can you explain what this is doing and why?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/ResidualProgramCompiler.kt
```diff
@@ -0,0 +1,754 @@
...
+    fun MethodVisitor.emitPrecompiledPluginsBlock(program: Program.Plugins) {
+
+        val precompiledPluginsBlock = compilePlugins(program)
+
+        precompiledScriptClassInstantiation(precompiledPluginsBlock) {
+
+            // val collector = PluginRequestCollector(scriptSource)
+            emitPluginRequestCollectorInstantiation()
+
+            // ${plugins}(temp.createSpec(lineNumber))
```
JLLeitschuh: :o: Do you want to leave this commented out code here?

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/ProjectSchemaAccessorsIntegrationTest.kt
```diff
@@ -70,6 +75,47 @@ class ProjectSchemaAccessorsIntegrationTest : AbstractIntegrationTest() {
         build("tasks")
     }
 
+    @Test
+    fun `conflicting extensions across build runs`() {
```
eskatos: 👍

subprojects/provider/provider.gradle.kts
```diff
@@ -67,11 +67,14 @@ val processResources by tasks.getting(ProcessResources::class) {
...
-        dependsOn(customInstallation)
-    }
+
+tasks.named("test").configure {
+    dependsOn(prepareIntegrationTestFixtures)
+    dependsOn(customInstallation)
+}
+
+tasks.register("quickTest", Test::class.java) {
+    exclude("**/*IntegrationTest.class")
```
JLLeitschuh: Is the `.class` necessary?

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/execution/ResidualProgramCompilerTest.kt
```diff
@@ -471,7 +485,8 @@ class ResidualProgramCompilerTest : TestWithTempFiles() {
                     scriptTemplateId,
                     sourceHash,
                     ProgramKind.TopLevel,
-                    ProgramTarget.Project)
+                    ProgramTarget.Project,
+                    accessorsClassPath)
```
JLLeitschuh: This method now has 9 parameters. It might be reaching the point where it would make sense to group them into an object.

bamboo: Generally I'd agree with you, @JLLeitschuh, in this concrete case however, that would mean memory indirections which is exactly the sort of thing I'm trying to avoid.

JLLeitschuh: I didn't realize that this code was that much of a "hot path" that you were trying to optimize for.
Fair enough.
👍

Issue #891, opened 2018-05-23
IDE Script Dependencies Resolver results collections should be stable


eskatos:
It happens that some of the returned collections don't ensure a stable order, making the IDE believe that new resolved dependencies are different than previous/cached ones.

The Gradle Kotlin DSL script dependencies resolver should guarantee stable ordering.

At least the `sources` collection was observed with changing order.

eskatos:
#893 fix the source path ordering stability
After more tests and scrutiny I'm confident other collections order is stable.
Closing as delivered

PR #893, opened 2018-05-23
Let script source dependencies order be stable

eskatos:
See https://github.com/gradle/kotlin-dsl/issues/891

### DIFF
--- a/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/SourcePathProvider.kt
+++ b/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/SourcePathProvider.kt
@@ -73,5 +73,5 @@ object SourcePathProvider {
 
 internal
 fun subDirsOf(dir: File): Collection<File> =
-    if (dir.isDirectory) dir.listFiles().filter { it.isDirectory }
+    if (dir.isDirectory) dir.listFiles().filter { it.isDirectory }.sortedBy { it.name }
     else emptyList()
--- a/subprojects/tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinLibSources.kt
+++ b/subprojects/tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinLibSources.kt
@@ -109,7 +109,8 @@ fun resolveSourcesUsing(dependencyHandler: DependencyHandler, query: ArtifactRes
             .resolvedComponents
             .flatMap { it.getArtifacts(SourcesArtifact::class.java) }
             .filterIsInstance<ResolvedArtifactResult>()
-            .map { it.file })
+            .map { it.file }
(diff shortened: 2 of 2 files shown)


subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/SourcePathProvider.kt
```diff
@@ -73,5 +73,5 @@ object SourcePathProvider {
 
 internal
 fun subDirsOf(dir: File): Collection<File> =
-    if (dir.isDirectory) dir.listFiles().filter { it.isDirectory }
+    if (dir.isDirectory) dir.listFiles().filter { it.isDirectory }.sortedBy { it.name }
```
bamboo: ❓ Would it be possible to sort only once at https://github.com/gradle/kotlin-dsl/blob/e035db702e25034569f117bad4fb0c7f87d1e71a/subprojects/tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinBuildScriptModelBuilder.kt#L305 ?

eskatos: That would fix the order instability too. But it would break the `gradleSource()` then `classpathSources` then `accessorsClassPath.src` order. Moreover, when some source is available more than once in the source path, the IDE usually present the first one (and allows to pick choose others). Sorting the whole thing once would also break the assumed same-as-compile-classpath order, potentially inducing surprising behavior. I also think that keeping the overall order makes it easier to troubleshoot issues if need be.

`classpathSources` is "unsorted" anyway, because of https://github.com/gradle/gradle/issues/5507. Sorting buckets that need to be sorted once would be a good tradeoff, I'll amend the PR in this regard.

subprojects/tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinLibSources.kt
```diff
@@ -109,7 +109,8 @@ fun resolveSourcesUsing(dependencyHandler: DependencyHandler, query: ArtifactRes
             .resolvedComponents
             .flatMap { it.getArtifacts(SourcesArtifact::class.java) }
             .filterIsInstance<ResolvedArtifactResult>()
-            .map { it.file })
+            .map { it.file }
+            .sorted()) // TODO remove sorting once https://github.com/gradle/gradle/issues/5507 is fixed
```
JLLeitschuh: Why do you sort by `it.name` above and just by the natural `File` sorting behavior here?

eskatos: Above is sorting child directories of a given directory.
This is a temporary workaround sorting a set of files coming out in unpredictable order from an `ArtifactResolutionQuery`.

PR #914, opened 2018-06-06
Generates Kotlin extensions for the Gradle API

eskatos:
See #221 

Maps:
- `Class<T>` to `KClass<T>`
- Groovy named arguments (`Map<String, ?>` as first parameter) to `vararg Pair<String, Any?>`


subprojects/build-plugins/build-plugins.gradle.kts
```diff
@@ -0,0 +1,55 @@
...
+
+val processResources: ProcessResources by tasks
+val writeKotlinDslProviderVersion by tasks.creating(WriteProperties::class) {
+    outputFile = processResources.destinationDir.resolve("${base.archivesBaseName}-versions.properties")
+    property("kotlin-dsl", version)
+}
+processResources.dependsOn(writeKotlinDslProviderVersion)
+
+
+val writeTestKitPluginClasspath by tasks.creating {
```
StefMa: Why not applying the [java gradle [dev] plugin](https://docs.gradle.org/current/userguide/java_gradle_plugin.html)? 🤔

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/plugins/GeneratedSourcesLayoutPlugin.kt
```diff
@@ -0,0 +1,66 @@
...
+import javax.inject.Inject
+
+
+internal
+class GeneratedSourcesLayoutPlugin : Plugin<Project> {
+
+    override fun apply(project: Project): Unit = project.run {
+
+        extensions.create(
+            "generatedSourcesLayout",
```
StefMa: Can we use named parameters here?

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/plugins/GeneratedSourcesLayoutPlugin.kt
```diff
@@ -0,0 +1,66 @@
...
+import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.provider.Provider
+import org.gradle.api.tasks.SourceSet
+
+import org.gradle.kotlin.dsl.*
+
+import javax.inject.Inject
+
+
+internal
```
StefMa: Should be open, right?
Or is internal "similar" to open?

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/plugins/KotlinDslJavaApiExtensionsPlugin.kt
```diff
@@ -0,0 +1,215 @@
...
+import org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+
+class KotlinDslJavaApiExtensionsPlugin : Plugin<Project> {
+
+    override fun apply(project: Project): Unit = project.run {
+
+        apply<GeneratedSourcesLayoutPlugin>()
+        apply(plugin = "kotlin")
```
StefMa: Where is the different of these two styles of applying plugins? 🤔

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/plugins/KotlinDslJavaApiExtensionsPlugin.kt
```diff
@@ -0,0 +1,215 @@
...
+                    from(extensionsSourceSet.output)
+                    dependsOn(extensionsSourceSet.classesTaskName)
+                }
+            }
+        }
+
+    private
+    fun Project.registerParameterNamesIndexTaskFor(
+        extensionsSet: KotlinDslApiExtensionsSet
+    ): TaskProvider<GenerateParameterNamesIndexProperties> =
```
StefMa: That looks a little bit odd.
Why not using the default `{` here?

subprojects/build-plugins/build-plugins.gradle.kts
JLLeitschuh: Can you add a description to this task to explain what it's doing and why?

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/plugins/GeneratedSourcesLayoutPlugin.kt
JLLeitschuh: It's not similar. `internal` is "similar" to package private in Java.

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/plugins/KotlinDslJavaApiExtensionsPlugin.kt
```diff
@@ -0,0 +1,215 @@
...
+                        freeCompilerArgs += "-Xjsr305=strict"
+                    }
+                }
+                getByName<Jar>("jar") {
+                    from(extensionsSourceSet.kotlin.sourceDirectories)
+                    from(extensionsSourceSet.output)
+                    dependsOn(extensionsSourceSet.classesTaskName)
+                }
+            }
+        }
```
JLLeitschuh: Do you also need to configure the "apielements" outgoing variants?
https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_known_issues
```
configurations {
    "apiElements" {
        outgoing
            .variants
            .getByName("classes")
            .artifact(mapOf(
                "file" to compileKotlin.destinationDir,
                "type" to "java-classes-directory",
                "builtBy" to compileKotlin
            ))
    }
}
```

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/plugins/KotlinDslJavaApiExtensionsPlugin.kt
```diff
@@ -0,0 +1,215 @@
...
+    val nameComponents
+        get() = project.run {
+            if (project == rootProject) listOf(project.name, sourceSet.get().name)
+            else listOf(rootProject.name, project.path, sourceSet.get().name)
+        }
+
+    internal
+    val sources = project.provider {
+        sourceSet.get().java.sourceDirectories
+            .map { project.fileTree(it) as FileTree }
```
JLLeitschuh: Why is this cast necessary?

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/tasks/GenerateKotlinDslApiExtensions.kt
```diff
@@ -0,0 +1,164 @@
...
+    @OutputDirectory
+    @PathSensitive(PathSensitivity.RELATIVE)
+    val outputDirectory = project.layout.directoryProperty()
+
+    @TaskAction
+    @Suppress("unused")
+    internal
+    fun generateKotlinDslApiExtensions() =
+        if (isUseEmbeddedKotlinDslProvider.get()) runEmbedded()
+        else runIsolated()
```
JLLeitschuh: What's the advantage of doing this? Doesn't the class-loader have to load the method `generateKotlinDslApiExtensionsSourceTo` either way simply because `GenerateKotlinDslApiExtensions` is loaded?

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/tasks/GenerateParameterNamesIndexProperties.kt
```diff
@@ -0,0 +1,118 @@
...
+
+        val qdoxBuilder = JavaProjectBuilder(sortedClassLibraryBuilderWithClassLoaderFor(classpath))
+        val qdoxSources = sources.asSequence().mapNotNull { qdoxBuilder.addSource(it) }
+
+        val properties = qdoxSources
+            .flatMap { it.classes.asSequence().filter { it.isPublic } }
+            .flatMap { it.methods.asSequence().filter { it.isPublic && it.parameterTypes.isNotEmpty() } }
+            .map { method ->
+                fullyQualifiedSignatureOf(method) to commaSeparatedParameterNamesOf(method)
+            }.toMap(linkedMapOf())
```
JLLeitschuh: Why was a `LinkedMap` chosen specifically here?

subprojects/build-plugins/src/test/kotlin/org/gradle/kotlin/dsl/build/plugins/GradleApiExtensionsIntegrationTest.kt
```diff
@@ -0,0 +1,189 @@
...
+
+            repositories {
+                jcenter()
+            }
+        """)
+
+        run("generateKotlinDslApiExtensions")
+
+        existing("build/generated-sources").walkTopDown().single { it.isFile }.let {
+            Assert.assertThat(it.name, CoreMatchers.equalTo("GeneratedGradleApiExtensionsMainKotlinDslApiExtensions.kt"))
```
JLLeitschuh: Are you intentionally using the non-static import version of `assertThat`?

subprojects/build-plugins/src/test/kotlin/org/gradle/kotlin/dsl/build/plugins/GradleApiExtensionsIntegrationTest.kt
```diff
@@ -0,0 +1,189 @@
...
+                    "org/gradle/kotlin/dsl/GeneratedGradleApiExtensionsMainKotlinDslApiExtensionsKt.class")
+                    .map { it.replace('/', File.separatorChar) }.toTypedArray()))
+        }
+    }
+}
+
+
+// See https://github.com/gradle/gradle/blob/master/buildSrc/subprojects/configuration/src/main/kotlin/org/gradle/gradlebuild/public-api.kt
+private
+val gradlePublicApiObjectDeclaration = """
```
JLLeitschuh: It would probably be good to have a reference back to this file in that one just incase the gradle/gradle one changes.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
```diff
@@ -0,0 +1,528 @@
...
+        "java.util.Collection" to "kotlin.collections.Collection",
+        "java.util.List" to "kotlin.collections.List",
+        "java.util.ArrayList" to "kotlin.collections.ArrayList",
+        "java.util.Set" to "kotlin.collections.Set",
+        "java.util.HashSet" to "kotlin.collections.HashSet",
+        "java.util.LinkedHashSet" to "kotlin.collections.LinkedHashSet",
+        "java.util.Map" to "kotlin.collections.Map",
+        "java.util.Map.Entry" to "kotlin.collections.Map.Entry",
+        "java.util.HashMap" to "kotlin.collections.HashMap",
+        "java.util.LinkedHashMap" to "kotlin.collections.LinkedHashMap")
```
JLLeitschuh: :o: I would swear that this same map already exists somewhere in the codebase somewhere. I remember reviewing a PR with it. Should this really be duplicated here?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/ClassBytesRepository.kt
```diff
@@ -0,0 +1,191 @@
...
+        && !endsWith("package-info$classFileExtension")
+        && !matches(compilerGeneratedClassFilePath)
+
+
+private
+const val classFileExtension = ".class"
+
+
+private
+val compilerGeneratedClassFilePath = Regex(".*\\$\\d+\\.class$")
```
JLLeitschuh: Probably a stupid question. Is there ever a generated class file with `$` but no number?

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/plugins/GeneratedSourcesLayoutPlugin.kt
StefMa: Then it should be open, right? 😂 
Otherwise Gradle can't create the Proxy-Class of this Plugin...

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/plugins/KotlinDslJavaApiExtensionsPlugin.kt
StefMa: I don't like the expression syntax if that kind.
I'do prefer to use:
```kotlin
fun Project.registerParameterNamesIndexTaskFor(
    extensionsSet: KotlinDslApiExtensionsSet
): TaskProvider<GenerateParameterNamesIndexProperties> {
// Source code
}
```

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/plugins/GeneratedSourcesLayoutPlugin.kt
eskatos: Plugin classes don't need to be open.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
eskatos: You did see it in #815 that was closed and not merged.

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/plugins/GeneratedSourcesLayoutPlugin.kt
eskatos: No we can't. This is a Java API which bytecode is built using Java 7 and doesn't include parameter names.

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/tasks/GenerateParameterNamesIndexProperties.kt
eskatos: To preverse ordering.

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/tasks/GenerateKotlinDslApiExtensions.kt
eskatos: `runEmbedded()` is useful only for development and testing purpose, and is not the default.
`runIsolated()` allows to use a version of the extension generation code different from the version of the embedded kotlin-dsl running the build where this plugin is applied.

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/tasks/GenerateParameterNamesIndexProperties.kt
JLLeitschuh: Might be good to mention that in a comment.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
JLLeitschuh: Okay, thanks.

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/plugins/KotlinDslJavaApiExtensionsPlugin.kt
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+    private
+    fun Project.registerParameterNamesIndexTaskFor(
+        extensionsSet: KotlinDslApiExtensionsSet
+    ): TaskProvider<GenerateParameterNamesIndexProperties> =
+
+        tasks.register(
+            extensionsSet.javaParameterNamesTaskName,
+            GenerateParameterNamesIndexProperties::class.java
```
JLLeitschuh: Is there a reified version of this function? If not, should there be one?

subprojects/build-plugins/src/main/kotlin/org/gradle/kotlin/dsl/build/plugins/KotlinDslJavaApiExtensionsPlugin.kt
```diff
@@ -0,0 +1,225 @@
...
+                it.classes.from(extensionsSet.sourceSetJavaOutputDir)
+                it.classpath.from(sourceSet.compileClasspath)
+                it.includes.set(extensionsSet.includes)
+                it.excludes.set(extensionsSet.excludes)
+                it.parameterNamesIndices.from(files(parameterNamesTasks.map { it.get() }))
+
+                it.outputDirectory.set(sourcesOutputDirFor(extensionsSourceSet, "kotlinDslExtensions"))
+
+                it.dependsOn(sourceSet.output)
+            }
```
JLLeitschuh: Do you also want to set a group?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
```diff
@@ -0,0 +1,528 @@
...
+ * - does not distinguish co-variance and contra-variance
+ */
+internal
+class ApiTypeProvider(
+    private val repository: ClassBytesRepository,
+    parameterNamesSupplier: ParameterNamesSupplier
+) : Closeable {
+
+    private
+    val context = Context(this, parameterNamesSupplier)
```
JLLeitschuh: Doesn't the compiler warn you about leaking a reference to `this` in the constructor here?
Couldn't this be solved with a `lazy` here?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
```diff
@@ -0,0 +1,528 @@
...
+    val sourceName: String,
+    private val delegateSupplier: () -> ClassNode,
+    private val context: ApiTypeProvider.Context
+) {
+
+    val isPublic: Boolean
+        get() = (ACC_PUBLIC and delegate.access) > 0
+
+    val isDeprecated: Boolean
+        get() = delegate.visibleAnnotations.has<java.lang.Deprecated>()
```
JLLeitschuh: You're only looking at Java code at this point correct?
I want to make sure you don't also need to look for a groovy `Deprecated` annotation or the Kotlin one.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
eskatos: Nope, `Context` is an inner static class, see below

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
eskatos: Yes, only Java bytecode

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiExtensionsGenerator.kt
```diff
@@ -329,15 +328,15 @@ fun List<ApiTypeUsage>.toTypeParametersString(type: ApiType? = null): String =
...
 private
-fun Bound.toKotlinString() =
+fun Variance.toKotlinString() =
     when (this) {
-        Bound.NONE -> ""
-        Bound.UPPER -> "out "
-        Bound.LOWER -> "in "
+        Variance.INVARIANT -> ""
+        Variance.COVARIANT -> "out "
+        Variance.CONTRAVARIANT -> "in "
```
JLLeitschuh: Why not put this directly on the enum itself?

eskatos: Because the enum sits in `ApiTypeProvider` which is about Java bytecode, not about Kotlin sources.

JLLeitschuh: Fair enough.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
```diff
@@ -0,0 +1,609 @@
...
+
+        fun ArrayDeque<String>.addSuperTypesOf(classNode: ClassNode) {
+            classNode.interfaces.forEach { push(it) }
+            if (classNode.superName != null) push(classNode.superName)
+        }
+
+        val stack = ArrayDeque<String>().apply {
+            addSuperTypesOf(delegate)
+        }
+        val visited = mutableListOf<String>()
```
bamboo: :x: `mutableSetOf<String>()`

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiTypeProvider.kt
```diff
@@ -0,0 +1,609 @@
...
+
+        val predicate = { candidate: MethodNode ->
+            candidate.desc == methodNode.desc && candidate.signature == methodNode.signature
+        }
+
+        while (stack.isNotEmpty()) {
+            val next = stack.pop()
+
+            if (next in visited) continue
```
bamboo: :x: `if (!visited.add(next)) continue`

subprojects/provider/src/test/java/org/gradle/kotlin/dsl/codegen/ClassAndGroovyNamedArguments.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package org.gradle.kotlin.dsl.codegen;
+
+import java.util.Map;
+import java.util.function.Consumer;
+
+
+public interface ClassAndGroovyNamedArguments {
```
bamboo: ⭕️ Consider moving these types used in tests to a sub-package or object.

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/codegen/GradleApiExtensionsIntegrationTest.kt
```diff
@@ -0,0 +1,202 @@
...
+                    println("$tag${'$'}{gradle.gradleUserHomeDir}")
+                }
+            }
+        """)
+
+        val gradleUserHomeDir = build("printGradleUserHomeDir", "-q").output.lines()
+            .single { it.contains(tag) }.substringAfter(tag).let { File(it) }
+
+        val generatedJar = gradleUserHomeDir.resolve("caches")
+            .listFiles { f -> f.isDirectory && f.name[0].isDigit() }.single()
```
bamboo: ⭕️ This looks very fragile and in fact, after a few iterations the `single()` call started failing.

subprojects/provider/src/test/java/org/gradle/kotlin/dsl/codegen/ClassAndGroovyNamedArguments.java
eskatos: Moved to `org.gradle.kotlin.dsl.fixtures.codegen`

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/codegen/GradleApiExtensionsIntegrationTest.kt
eskatos: Should be much more solid now

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiExtensionsGenerator.kt
```diff
@@ -18,39 +18,446 @@
...
+    const val gradleTypeOf = "org.gradle.api.reflect.TypeOf"
+    const val kotlinClass = "kotlin.reflect.KClass"
+    const val kotlinArray = "kotlin.Array"
+    const val kotlinCollection = "kotlin.collections.Collection"
+}
+
+
+private
+val ApiTypeUsage.isSAM
+    get() = type?.isSAM ?: isGradleAction
```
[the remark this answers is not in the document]
eskatos: Good catch, this is a leftover, removed

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiExtensionsGenerator.kt
```diff
@@ -18,39 +18,446 @@
...
+    parameters.map { MappedApiFunctionParameter(it) }
+
+
+private
+data class MappedApiFunctionParameter(
+    val original: ApiFunctionParameter,
+    val index: Int = original.index,
+    val type: ApiTypeUsage = original.type,
+    val isVarargs: Boolean = original.isVarargs,
+    val invocation: String = "${if (original.isVarargs) "*" else ""}`${original.name ?: "p$index"}`"
```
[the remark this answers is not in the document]
eskatos: I like it better 👍 renamed.

samples/model-rules/build.gradle.kts
```diff
@@ -1,2 +1,2 @@
 
-apply(mapOf("plugin" to my.Rules::class.java))
+apply("plugin" to my.Rules::class.java)
```
bamboo: This one is tricky because passing a `KClass` would not work. Not a huge deal.

PR #933, opened 2018-06-21
Decouple the versions of the Kotlin libraries used by the build and by the runtime

bamboo:
So that we can compile against a newer version of the Kotlin
libraries without incurring the risk of loading incompatible library
versions (like older plugins running against newer compilers, etc).

This makes upgrading to Kotlin 1.2.50 smoother.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -9,19 +9,12 @@ buildscript {
 
     build.loadExtraPropertiesOf(project)
 
-    val kotlinVersion: String by extra
-    dependencies {
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion")
-    }
-
-    repositories {
-        gradlePluginPortal()
-    }
 }
 
 plugins {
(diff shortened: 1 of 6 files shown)


build.gradle.kts
```diff
@@ -9,19 +9,12 @@ buildscript {
...
-
-    repositories {
-        gradlePluginPortal()
-    }
 }
 
 plugins {
     base
     id("org.jetbrains.gradle.plugin.idea-ext") version "0.1"
+    kotlin("jvm") apply false
```
eskatos: :o: I'd move this one line up, just for clarity and to prevent any ordering issue

bamboo: Ordering issue?

eskatos: There shouldn't be any between these two plugins.

eskatos: I'd just move the kotlin language plugin up the IDE one, so the list of plugins order reads nicer.

PR #944, opened 2018-06-28
Serialize TAPI model requests

bamboo:
In order to avoid spawning a Gradle daemon per open .gradle.kts script.


subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/concurrent/EventLoop.kt
```diff
@@ -0,0 +1,64 @@
...
+    fun ensureAliveConsumer() = synchronized(this) {
+        if (consumer?.isAlive != true) {
+            consumer = newConsumerThread()
+        }
+    }
+
+    private
+    fun newConsumerThread() = thread {
+        loop(::poll)
+    }
```
JLLeitschuh: Daemon thread? Also, name it for debugging?

PR #953, opened 2018-07-06
Fix Travis build

SUPERCILEX:
It looks pretty bad for every commit to fail CI. 😄


JLLeitschuh:
@SUPERCILEX THANK YOU!!! 😄

Issue #964, opened 2018-07-12
Cannot apply JacocoPlugin by type


ldaley:
### Expected Behavior

```
project.apply<JacocoPlugin>()
```

Should work.

### Current Behavior

Compiler complains with…
![image](https://user-images.githubusercontent.com/17825/42664755-c8623e56-867f-11e8-9523-7d646a722750.png)

Caused by:

```
public class JacocoPlugin implements Plugin<ProjectInternal>
```

I am rather uneducated on Kotlin's type system capabilities, but I can't see a way to make this work. Seems like Jacoco plugin should be changed to `implements Plugin<Project>` and do a cast in its implementation.

ldaley:
Workaround:

```
project.pluginManager.apply(JacocoPlugin::class.java)
```

eskatos:
This is one incarnation of the Gradle API leaking internal types.
I just checked and looots of core plugins do that.

I opened https://github.com/gradle/gradle/issues/5964 for fixing the plugins api.

If fixing the Gradle public API turns out to be difficult, we could consider changing the `apply<T>()` function signature on the kotlin-dsl side of things to:
```kotlin
inline fun <reified T : Plugin<out Project>> Project.apply()
```

big-guy:
What's the downside of `inline fun <reified T : Plugin<out Project>> Project.apply()`?

StefMa:
Regardless of the downside. I think you - as Gradle - should provide "best
practices" plugins. And providing a plugin with internal apis is by far not
a best practice... No? 🤔

bamboo:
@big-guy The downside would be the types promising more than they should causing code like `project.apply<SomeProjectSubtypePlugin>()` to succeed no matter what the subtype of `Project` the plugin expects, i.e., the code below would compile successfully but fail at runtime since Gradle's `Project` implementation doesn't implement `RandomProjectSubtype`:

```kotlin
class RandomProjectSubtypePlugin : org.gradle.api.Plugin<RandomProjectSubtype> {
  fun apply(target: RandomProjectSubtype) = Unit
}

interface RandomProjectSubtype : org.gradle.api.Project

project.apply<RandomProjectSubtypePlugin>()
```

eskatos:
Another way is to apply plugins by id:
`apply(plugin = "jacoco")`
This is the preferred way for core plugins.

Closing as superseded by gradle/gradle#5964

PR #965, opened 2018-07-13
Allow plugins which support only subtypes of project

JakeWharton:
The most notable one is Gradle's built-in `JavaPlugin` which implements `Plugin<ProjectInternal>`.


StefMa:
See #964
I think that should be discussed first...

JakeWharton:
Ah, yep. I didn't see that issue because I only searched for JavaPlugin.

eskatos:
Another way is to apply plugins by id:
`apply(plugin = "java")`
This is the preferred way for core plugins.

See #964 

Closing as superseded by gradle/gradle#5964

Issue #835, opened 2018-04-23
Fix User Guide links in builtin Gradle plugin spec accessors


bamboo:
The User Guide link on `java-gradle-plugin` points to a [non-existent page](https://docs.gradle.org/current/userguide/javaGradle_plugin.html).

We should review them all and put some tests in place.

PR #968, opened 2018-07-16
Fix and test generated kdoc links for core plugins

eskatos:
See #835


subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/codegen/UserGuideLinkTest.kt
```diff
@@ -1,45 +1,63 @@
...
 
 import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
+
+import java.net.HttpURLConnection
+import java.net.URL
+
+
+val linkedPlugins =
```
bamboo: ⭕️ I'd move the test classes up and these constants down.

PR #969, opened 2018-07-16
Let the embedded-kotlin and kotlin-dsl plugins warn on different kotlin versions

eskatos:


subprojects/plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/embedded/EmbeddedKotlinPlugin.kt
```diff
@@ -60,3 +65,18 @@ open class EmbeddedKotlinPlugin @Inject internal constructor(
...
     }
 }
+
+
+internal
+fun Logger.warnOnDifferentKotlinVersion(kotlinVersion: String?) {
+    if (kotlinVersion != embeddedKotlinVersion) {
+        warn(
+            """
+            WARNING: Unsupported Kotlin plugin version.
```
bamboo: :x: Indent the message inside the string for clarity.

Issue #962, opened 2018-07-12
Change in buildSrc is not easily propagated to IntelliJ build scripts editor


eskatos:
When changing something in `buildSrc`, the script dependencies for build scripts aren't updated, making the change not visible to the IntelliJ build script editor.

### Expected Behavior
Changes should be available right away.

### Current Behavior
Changes aren't propagated correctly.

This is because the Gradle Kotlin DSL script dependencies resolver does short-circuit dependencies requests that didn't see a change in either the `plugins` or `buildscript` block of scripts. This drastic short-circuit should be removed, for correctness. A more fine grained short-circuit mechanism could be introduced to compensate for the extra processing.

One workaround is to make a change in the `plugins` or `buildscript` block of the edited script.

### Context
https://youtrack.jetbrains.com/issue/KT-25354

JLLeitschuh:
@eskatos I'm still seeing this issue in my builds.
However, repeatedly refreshing or changing the `plugins` doesn't fix the issue for me. I have to restart IntelliJ to fix the issue.

neVERberleRfellerER:
This bug is indeed still present. With Versions.kt in buildSrc that contains dependency versions, classpath still contains old versions after versions were changed in Versions.kt even through gradle itself have already downloaded new dependencies. Restart or change to settings.gradle.kts of root project is necessary for IDEA to detect the changes.

PR #972, opened 2018-07-17
Let buildSrc changes be reflected in IntelliJ script editor instantly

eskatos:
by removing a pessimistic short-circuit in the script dependencies resolver.

See #962 

intellij-gradle kts-buildsrc-change-auto-reload

Note that with IntelliJ Kotlin Plugin <=1.2.60-eap-44 you need to explicitly save for it to work. A fix is in the works so that changing focus is enough.


eskatos:
@JLLeitschuh the removed test was asserting the now removed short-circuit.

Issue #907, opened 2018-06-04
Let container scope string invoke be `named`


bamboo:
The string invoke syntax on containers provides the shortest idiom for configuring elements but Its `maybeCreate` semantics is confusing and error-prone.

Considering that _configuring new elements_  is more common than _registering existing elements_  and that the `register(...) { ... }` syntax provides a convenient and clear replacement for the register  element_ scenario, the string invoke syntax semantics should be changed to `named`.

This requires the lazy API to be moved up in the Gradle container hierarchy: gradle/gradle-native#680.

gildor:
I disagree that "registering new elements is more common than configuring existing elements".
It's true for any kind DSL, except one very important: tasks.
Task configuration is a very common thing and uses invoke syntax. The situation can be improved if plugins migrate their configs from tasks to extensions, but now even Kotlin plugin or core Gradle plugins have a lot of features configured only by task configuration
But a big problem with tasks configuration is it's really ugly because for tasks you usually need type and invoke syntax doesn't support reified generics.
Maybe make sense to deprecate this syntax for tasks and be explicit and use `named`, `create` etc functions instead.

mkobit:
An additional consideration that I have seen with the string invocation DSL is unintentional creation when configuration was desired. For example

```kotlin
tasks {
  "asemble" {
    dependsOn(myTask)
  }
}
```

A typo of `assemble` to `assemble` doesn't lead to an error message here. A `maybeCreate` with the task is created and if the tasks aren't inspected closely it could be difficult to catch that your intended dependencies did not execute during the build.

gildor:
@mkobit really good point why implicit creation of tasks is bad practice 👍

eskatos:
Thanks for the feedback @gildor and @mkobit!
After discussing this we changed to let container scope string invoke be `named`.

PR #986, opened 2018-07-28
Let container scope string invoke be `named`

eskatos:
See #907

### DIFF
--- a/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectContainerExtensions.kt
+++ b/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectContainerExtensions.kt
@@ -69,7 +69,7 @@ class NamedDomainObjectContainerScope<T : Any>(
      * @see [NamedDomainObjectContainer.named]
      */
     operator fun String.invoke(configuration: T.() -> Unit): DomainObjectProvider<T> =
-        this().also { it.configure(configuration) }
+        this().apply { configure(configuration) }
 
     /**
      * @see [NamedDomainObjectContainer.named]
@@ -81,7 +81,7 @@ class NamedDomainObjectContainerScope<T : Any>(
      * @see [PolymorphicDomainObjectContainer.named]
      */
     operator fun <U : T> String.invoke(type: KClass<U>, configuration: U.() -> Unit): DomainObjectProvider<U> =
-        this(type).also { it.configure(configuration) }
+        this(type).apply { configure(configuration) }
 
     /**
      * @see [PolymorphicDomainObjectContainer.named]
--- a/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectContainerExtensions.kt
+++ b/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectContainerExtensions.kt
@@ -21,8 +21,12 @@ import org.gradle.api.DomainObjectProvider
 import org.gradle.api.NamedDomainObjectContainer
 import org.gradle.api.PolymorphicDomainObjectContainer
 
+import org.gradle.kotlin.dsl.support.illegalElementType
+import org.gradle.kotlin.dsl.support.uncheckedCast
+
 import kotlin.reflect.KClass
 import kotlin.reflect.KProperty
+import kotlin.reflect.full.safeCast
 
 
 /**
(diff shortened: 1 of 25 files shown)


subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectContainerExtensions.kt
```diff
@@ -66,28 +66,28 @@ class NamedDomainObjectContainerScope<T : Any>(
         polymorphicDomainObjectContainer().containerWithType(type)
 
     /**
-     * @see [NamedDomainObjectContainer.maybeCreate]
+     * @see [NamedDomainObjectContainer.named]
      */
-    inline operator fun String.invoke(configuration: T.() -> Unit): T =
-        this().apply(configuration)
+    operator fun String.invoke(configuration: T.() -> Unit): DomainObjectProvider<T> =
+        this().also { it.configure(configuration) }
```
bamboo: ⭕️ Why not `this().apply { configure(configuration) }`?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectContainerExtensions.kt
```diff
@@ -66,28 +66,28 @@ class NamedDomainObjectContainerScope<T : Any>(
...
+        this(type).also { it.configure(configuration) }
 
     /**
-     * @see [PolymorphicDomainObjectContainer.maybeCreate]
+     * @see [PolymorphicDomainObjectContainer.named]
      */
-    operator fun <U : T> String.invoke(type: KClass<U>): U =
-        polymorphicDomainObjectContainer().maybeCreate(this, type.java)
+    operator fun <U : T> String.invoke(type: KClass<U>): DomainObjectProvider<U> =
+        polymorphicDomainObjectContainer().withType(type.java).named(this)
```
bamboo: :x: The semantics should be that the caller is asserting the type of the named element, not querying elements of the specific type, similar to [`NamedDomainobjectCollection.getByName<T>(String)`](https://github.com/gradle/kotlin-dsl/blob/448edf838543abb886968bab59e9bc28f1c6111b/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectCollectionExtensions.kt#L60).

eskatos: The assertion would need to be lazy to not realize the container element, as a configuration block maybe. Then the assertion happens sometime in the future, when/if the element is realized.

eskatos: I agree the semantics are different. If the targeted element cannot be found (by type and name) you'll get a "not found" exception. What you'd not get is the detail about an element found with the requested name but of a different type.

bamboo: Yes, the assertion should happen during configuration.

eskatos: Then it requires an eager cast to return the requested provider type that may fail before we have a chance to provide a helpful error:
```kotlin
operator fun <U : T> String.invoke(type: KClass<U>): DomainObjectProvider<U> =
    polymorphicDomainObjectContainer().named(this).apply {
        configure {
            it as? U
                ?: illegalElementType(container, this@invoke, type, it::class)
        }
    } as DomainObjectProvider<U>
```

Or did I miss something?

bamboo: I was thinking on something like:

```kotlin
    /**
     * @see [PolymorphicDomainObjectContainer.named]
     */
    operator fun <U : T> String.invoke(type: KClass<U>, configuration: U.() -> Unit): DomainObjectProvider<U> =
        uncheckedCast(
            polymorphicDomainObjectContainer().named(this).apply {
                configure {
                    configuration(
                        type.safeCast(it)
                            ?: throw illegalElementType(container, name, type, it::class)
                    )
                }
            })
```

eskatos: Yeah, it's basically the same.
Wouldn't the unchecked cast fail earlier than the type check?

PR #987, opened 2018-07-28
Return build cache metadata from script caching commands

bamboo:
The cache commands will store and load `OriginTaskExecutionMetadata` as currently expected by the build-scan plugin.

Unfortunately this breaks the build-scan plugin assumption that a cache command is only ever triggered in the context of an executing task and will require a new plugin version before it can be used (and the corresponding integration test ignored).

### DIFF
--- a/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptEvaluator.kt
+++ b/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptEvaluator.kt
@@ -34,9 +34,13 @@ import org.gradle.internal.classloader.ClasspathHasher
 import org.gradle.internal.classpath.ClassPath
 import org.gradle.internal.classpath.DefaultClassPath
 import org.gradle.internal.hash.HashCode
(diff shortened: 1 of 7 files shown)


subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptEvaluator.kt
```diff
@@ -227,32 +231,52 @@ class StandardKotlinScriptEvaluator(
...
+                val cacheDir = File(baseDir, "cache").apply { require(mkdir()) }
+                initializeCacheDir(cacheDir, cacheKey, scriptHost, initializer)
+            }.resolve("cache")
+
+        private
+        fun initializeCacheDir(cacheDir: File, cacheKey: String, scriptHost: KotlinScriptHost<*>, initializer: (File) -> Unit) {
 
-                val cacheDir =
-                    File(baseDir, "cache").apply { require(mkdir()) }
+            // TODO: Move BuildCacheController integration to ScriptCache
```
eskatos: Not blocking merging, but this should be done early as a follow up

Issue #275, opened 2017-02-23
Write missing API documentation


bamboo:

eskatos:
All public api members of `gradle/kotlin-dsl` now have proper kdoc.

PR #988, opened 2018-07-30
Add a bunch of missing kdoc on Kotlin DSL public APIs

eskatos:
See #275

### DIFF
--- a/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/ConventionExtensions.kt
+++ b/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/ConventionExtensions.kt
@@ -36,18 +36,50 @@ inline fun <reified T : Any> Convention.getPluginByName(name: String): T =
     } ?: throw IllegalStateException("A convention named '$name' could not be found.")
 
 
-inline fun <reified T : Any> Convention.getPlugin() =
+/**
+ * Locates the plugin convention object with the given type.
+ *
+ * @param T the convention plugin type
+ * @return the convention plugin
+ * @throws [IllegalStateException] when there is no such object contained in this convention, or when there are multiple such objects
+ * @see [Convention.getPlugin]
+ */
+inline fun <reified T : Any> Convention.getPlugin(): T =
     getPlugin(T::class)
 
 
+/**
+ * Locates the plugin convention object with the given type.
+ *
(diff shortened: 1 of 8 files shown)


subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/ConventionExtensions.kt
```diff
@@ -36,18 +36,50 @@ inline fun <reified T : Any> Convention.getPluginByName(name: String): T =
...
+inline fun <reified T : Any> Convention.getPlugin(): T =
     getPlugin(T::class)
 
 
+/**
+ * Locates the plugin convention object with the given type.
+ *
+ * @param conventionType the convention plugin type
+ * @return the convention plugin
+ * @throws [IllegalStateException] when there is no such object contained in this convention, or when there are multiple such objects
```
JLLeitschuh: Are you sure this is `IllegalStateException`? Sounds like it should be an `IllegalArgumentException`.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/DependencyHandlerScope.kt
```diff
@@ -194,6 +194,9 @@ class DependencyHandlerScope(val dependencies: DependencyHandler) : DependencyHa
     inline operator fun <T : ModuleDependency> Configuration.invoke(dependency: T, dependencyConfiguration: T.() -> Unit): T =
         add(name, dependency, dependencyConfiguration)
 
+    /**
+     * `dependencies` block.
```
JLLeitschuh: This line of documentation isn't all that helpful.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/ConventionExtensions.kt
eskatos: Yes, it is, see https://github.com/gradle/gradle/blob/master/subprojects/core-api/src/main/java/org/gradle/api/plugins/Convention.java#L46

JLLeitschuh: Strange, not the exception I would have expected.

Issue #991, opened 2018-07-31
Remove usage of ProgressLogger.setLoggingHeader


oehme:
This method creates verbose, persistent log output. It is also deprecated and I'd like to remove it. See https://github.com/gradle/gradle/pull/6104. Instead, wrap a build operation around the JAR generation (in case you don't have that already) to get proper rendering in the rich console progress area.

PR #992, opened 2018-07-31
Replace deprecated ProgressLogger.setLoggingHeader by build operation

eskatos:
See #991 

generated-jar-progress

The line gets quite long.


eskatos:
@bamboo, I added coverage for the build operation descriptor

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptClassPathProvider.kt
```diff
@@ -161,9 +166,22 @@ class KotlinScriptClassPathProvider(
     private
     fun produceFrom(id: String, generate: JarGeneratorWithProgress): File =
         jarCache(id) { outputFile ->
-            progressMonitorFor(outputFile, 3).use { progressMonitor ->
-                generateAtomically(outputFile) { generate(it, progressMonitor::onProgress) }
-            }
+            buildOperationExecutor.run(object : RunnableBuildOperation {
```
bamboo: ❓ Do we need both? The progress monitor and the build operation?

eskatos: afaik yes, @oehme can you confirm?

Issue #464, opened 2017-08-16
KT-22210 Use meaningful parameter names instead of `p0`


sdeleuze:
In IntelliJ IDEA, when you enable the option that displays implicit parameter names, you get weird `p0` parameters in `id()` and `plugin()`. Meaningful parameter names would be better.

![screenshot from 2017-08-16 14-39-13](https://user-images.githubusercontent.com/141109/29364462-c7cb2e3a-8293-11e7-8fd1-f4fdcc738dc5.png)

JLLeitschuh:
Is this because gradle is compiled without the `-parameters` compiler flag? This sounds like something that needs to be fixed in the core gradle repository.

https://docs.oracle.com/javase/tutorial/reflect/member/methodparameterreflection.html

eskatos:
> https://github.com/gradle/kotlin-dsl/issues/437#issuecomment-317001913
> Upon further inspection this seems to be caused by Kotlin interface delegation (`... : Project by project { ...`) not preserving the parameter names in the generated bytecode.
>
> Parameter names show up correctly for non delegated members, for instance, notice the version parameter below:
![screen shot 2017-07-21 at 10 32 31 am](https://user-images.githubusercontent.com/51689/28465674-fe24a6da-6dff-11e7-9d67-d0001099da38.png)

AlexCzar:
@eskatos I searched Kotlin's issue tracker but couldn't find this specific problem. Was it reported?

eskatos:
@AlexCzar not that I know of, feel free to report it upstream, I'll get back to it next week if you don't beat me to it.

It doesn't happen when no `.kts` file is involved.
I didn't try and reproduce outside of the Kotlin DSL yet.

eskatos:
Reported upstream: https://youtrack.jetbrains.com/issue/KT-22210

bamboo:
Moved to backlog so we can decide if we should tackle it ourselves.

eskatos:
Two candidate solutions until Gradle is built with Java 8 and `-parameters`:
- have a parameter names index in the Gradle distribution and enhance the Gradle API Jar before passing it to the Kotlin compiler in the Kotlin DSL build
- building all Gradle modules with Java 8 and `-parameters`, publish them and use them instead of `gradleApi()` in the Kotlin DSL build

bamboo:
Tracked it down to https://youtrack.jetbrains.com/issue/KT-25193

JLLeitschuh:
[...]

PR #1004, opened 2018-08-04
Fix missing parameter names in Gradle Kotlin DSL jars

eskatos:
Functions on types using kotlin [implementation by delegation](https://kotlinlang.org/docs/reference/delegation.html#implementation-by-delegation) for a Java interface miss parameter names metadata.

See #464 

Reusing the `gradle-api-metadata`, an `ArtifactTransform` hydrates the Gradle API bytecode provided by `gradleApi()` with Java 8 parameter names. The Gradle Kotlin DSL is compiled against the transformed Gradle API disabling the kotlin compiler fast class reader. This combination lets the parameter names show up properly in the IDE instead of `p0`, `p1` ...

image

### DIFF
--- a/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinBuildScript.kt
+++ b/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinBuildScript.kt
@@ -46,7 +46,6 @@ annotation class KotlinScriptTemplate
     resolver = KotlinBuildScriptDependenciesResolver::class,
     scriptFilePattern = ".*\\.gradle\\.kts")
 @ScriptTemplateAdditionalCompilerArguments([
-    "-Xuse-old-class-files-reading",
     "-Xjsr305=strict",
     "-Xprogressive",
     "-XXLanguage:+NewInference",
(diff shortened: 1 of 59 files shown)


subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/KotlinBuildScript.kt
```diff
@@ -46,6 +46,8 @@ annotation class KotlinScriptTemplate
     resolver = KotlinBuildScriptDependenciesResolver::class,
     scriptFilePattern = ".*\\.gradle\\.kts")
 @ScriptTemplateAdditionalCompilerArguments([
+    "-java-parameters",
```
bamboo: ❓ I don't understand why the new compiler options are also required in the script templates. I understood `-Xuse-old-class-files-reading` was required to get the compiler to load parameter names from the transformed Gradle api jar and `-java-parameters` was about preserving parameter names in the emitted class files, none of which seem to apply to the scripts, what am I missing?

eskatos: Right. I've been overzealous. We don't need those for script compilation. And we don't need `-Xuse-old-class-files-reading` on the `kotlin-dsl` plugin.

PR #1006, opened 2018-08-04
Initial support for script compile build operations

breskeby:
### Context
We want to capture script compile for kotlin scripts in gradle 4.10.  I added simple func test coverage.

### Contributor Checklist


bamboo:
Thanks, @breskeby, we'll take it from here unless you have pending changes?

breskeby:
Thank you. No pending changes from my side 

cheers,
rené

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/execution/ResidualProgramCompiler.kt
```diff
@@ -213,7 +218,7 @@ class ResidualProgramCompiler(
 
     fun emitStage2ProgramFor(scriptFile: File, originalPath: String) {
 
-        val precompiledScriptClass = compileScript(scriptFile, originalPath, stage2ScriptDefinition)
+        val precompiledScriptClass = compileScript(scriptFile, originalPath, stage2ScriptDefinition, "BODY")
```
ldaley: @bamboo can we get something put around these constants to make it clearer that they should remain stable as much as possible? If the do change, we'll just need to be made aware.

We will know from test breakages, but it would be good for later readers to understand that these are consequential.

bamboo: Makes sense, @ldaley. Done in c42c710.

PR #1007, opened 2018-08-05
 Let all plugins have a plugin id

eskatos:
Before https://scans.gradle.com/s/ee5gn4c6wld4i/plugins

image

After https://scans.gradle.com/s/s4qeko73toyfu/plugins

image


subprojects/provider-plugins/src/main/resources/META-INF/gradle-plugins/org.gradle.kotlin.kotlin-dsl-project.properties
```diff
@@ -0,0 +1 @@
+implementation-class=org.gradle.kotlin.dsl.provider.plugins.KotlinScriptBasePlugin
```
bamboo: ❓ Does this mean we'll get plugin spec accessors for these plugins too?

eskatos: Nope, only the gradle jars are considered when generating the plugin id extensions.

PR #1009, opened 2018-08-06
Introduce the `existing` and `registering` property delegates

bamboo:


vlsi:
Just wondering: could `tasks.registering<Zip>` be added as well?

For instance:

```
val distZip by tasks.registering<Zip> {
  ...
}
```

instead of the current

```
val distZip by tasks.registering(Zip::class) {
  ...
}
```

Issue #390, opened 2017-06-08
kotlin-dsl script compilation and plugin apply the all-open kotlin compiler plugin for Gradle types


eskatos:
all-open Gradle types (Task..)

eskatos:
The `allopen` kotlin compiler plugin only allows to use annotations to tag types. See https://kotlinlang.org/docs/reference/compiler-plugins.html

We'll either need to add such an annotation to the Gradle API and targeted types, or, praise for adding support for listing open types in the `allopen` kotlin compiler plugin configuration.

jmfayard:
I was to quick https://github.com/gradle/kotlin-dsl/issues/1189

jmfayard:
More infos about allopen: 

> If the class (or any of its superclasses) is annotated with com.your.Annotation, the class itself and all its members will become open. It even works with meta-annotations:

https://github.com/Kotlin/KEEP/blob/master/proposals/extensions/opening-classes-for-frameworks.md

https://blog.jetbrains.com/kotlin/2016/12/kotlin-1-0-6-is-here/

eskatos:
@jmfayard, it looks great!
It's already too late for 5.0. It's now tentatively scheduled for 5.1.

jmfayard:
Ok thanks Paul !

mkobit:
As an alternative, why doesn't Gradle create proxy classes for `final` classes similar to how ByteBuddy does for Mockito?

JLLeitschuh:
@mkobit I think that that's considered a pretty expensive operation. I'm guessing there are other reasons as well.

eskatos:
A prototype of `allopen` integration is available in the [eskatos/compilation/allopen](https://github.com/gradle/kotlin-dsl/compare/develop...eskatos/compilation/allopen) branch. It's not satisfactory yet in that the compiler plugins api for embedded compiler usage are only available in the maven plugin artifact.

Also see the discussion in https://github.com/gradle/gradle/pull/7971

PR #1024, opened 2018-08-10
Refine task-dependencies sample

eskatos:


samples/task-dependencies/build.gradle.kts
```diff
@@ -1,21 +1,47 @@
+/**
+ * An example Gradle task.
+ */
+open class GreetingTask : DefaultTask() {
```
JLLeitschuh: Has the `all-open` plugin not been applied to build scripts yet? Is there still a plan to do so?

StefMa: No, its not.
See https://github.com/gradle/kotlin-dsl/issues/390

Issue #1034, opened 2018-08-13
NamedDomainObjectCollection.withType<T>() is missing


eskatos:
### Expected Behavior
`tasks.withType<GroovyCompile>().named("compileGroovy")` should compile

### Current Behavior
 If I try to use `tasks.withType<GroovyCompile>().named("compileGroovy")` this fails because the `withType` extension is only for `DomainObjectCollection` and _named_ domain object containers override the return type of `withType` to return a `NamedDomainObjectCollection`.

PR #1035, opened 2018-08-13
Add missing specialized withType<T>() extensions on NamedDomainObjectCollection

eskatos:
See #1034

### DIFF
--- a/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectCollectionExtensions.kt
+++ b/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectCollectionExtensions.kt
@@ -29,6 +29,34 @@ import kotlin.reflect.full.safeCast
 
 
 /**
+ * Returns a collection containing the objects in this collection of the given type. Equivalent to calling
+ * {@code withType(type).all(configureAction)}
+ *
+ * @param S The type of objects to find.
+ * @param configuration The action to execute for each object in the resulting collection.
+ * @return The matching objects. Returns an empty collection if there are no such objects
+ * in this collection.
+ * @see [NamedDomainObjectCollection.withType]
+ */
+inline fun <reified S : Any> NamedDomainObjectCollection<in S>.withType(noinline configuration: S.() -> Unit): NamedDomainObjectCollection<S> =
+    withType().also { it.all(configuration) }
+
+
+/**
+ * Returns a collection containing the objects in this collection of the given type. The
+ * returned collection is live, so that when matching objects are later added to this
+ * collection, they are also visible in the filtered collection.
+ *
+ * @param S The type of objects to find.
+ * @return The matching objects. Returns an empty collection if there are no such objects
+ * in this collection.
+ * @see [NamedDomainObjectCollection.withType]
+ */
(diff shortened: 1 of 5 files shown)


subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/NamedDomainObjectCollectionExtensions.kt
```diff
@@ -28,6 +28,34 @@ import kotlin.reflect.KProperty
...
+ * {@code withType(type).all(configureAction)}
+ *
+ * @param S The type of objects to find.
+ * @param configuration The action to execute for each object in the resulting collection.
+ * @return The matching objects. Returns an empty collection if there are no such objects
+ * in this collection.
+ * @see [NamedDomainObjectCollection.withType]
+ */
+inline fun <reified S : Any> NamedDomainObjectCollection<in S>.withType(noinline configuration: S.() -> Unit): NamedDomainObjectCollection<S> =
+    withType().also { it.all(configuration) }
```
JLLeitschuh: Why not just expose an `invoke` method on `NamedDomainObjectCollection` instead?

bamboo: :x: `configureEach`?

JLLeitschuh: @bamboo Not in this case. `withType` that takes an `Action` always uses the immediate configuration API (calling `all`).

This is the docs from `DomainObjectCollection`:
```java
    /**
     * Returns a collection containing the objects in this collection of the given type. Equivalent to calling
     * {@code withType(type).all(configureAction)}
     *
     * @param type The type of objects to find.
     * @param configureAction The action to execute for each object in the resulting collection.
     * @return The matching objects. Returns an empty collection if there are no such objects in this collection.
     */
    <S extends T> DomainObjectCollection<S> withType(Class<S> type, Action<? super S> configureAction);
```

bamboo: :x: Right. It would be inconsistent to deviate from that.

I would remove the extension until a similar method exists in the core Gradle API.

JLLeitschuh: @bamboo Can you elaborate on why you think this? Maybe I'm simply misunderstanding you.

eskatos: Makes sense to remove it for now.
`NamedDomainObjectCollection` only overrides `withType(Class)` from `DomainObjectCollection`

bamboo: @JLLeitschuh For two reasons:
1. the DSL sugar should encourage configuration avoidance whenever possible
2. `NamedDomainObjectCollection` only overrides `withType(Class)` not `withType(Class, Action)`

Issue #1032, opened 2018-08-12
Use of project property as opt-in toggle for build caching of scripts is cumbersome in practice


ldaley:
### Expected Behavior

A convenient and idiomatic way to opt-in to this functionality on a per user or per project basis.

### Current Behavior

I can either explicitly add the project property to the invocation, which is cumbersome, or use some obtuse `ORG_GRADLE_PROJECT_org.gradle.kotlin.dsl.caching.buildcache=true` env var.

I should have the same options for enabling this as I do for build cache (probably minus the first class invocation option).  That is, I should be able to put something in my «GRADLE_USER_HOME»/gradle.properties or the same file in the project root, as I can with `org.gradle.caching`.

ldaley:
If we do change this, also need to update here: https://github.com/gradle/gradle/blob/4d620bab0fdecbc990ebc8c02775ea230578d06a/subprojects/core/src/main/java/org/gradle/internal/scan/config/BuildScanPluginCompatibility.java#L69

eskatos:
The following in a `gradle.properties` file should work already:

```properties
org.gradle.project.org.gradle.kotlin.dsl.caching.buildcache=true
```

ldaley:
@eskatos if you can turn that `should` into `does`, then we can close this.

eskatos:
@ldaley it doesn't work because the property is checked on `StartParameter` which doesn't resolve properties from `gradle.properties` 

https://github.com/gradle/kotlin-dsl/blob/b9cb04969dab25fdf549de5b32231fc6326883f4/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/cache/BuildServices.kt#L49-L50

This needs to be fixed.

eskatos:
We should not make that flag a public build option. Its purpose is to test drive script compilation caching during the RC phase. The plan is to get rid of it and simply tie enabling script compilation caching to enabling the build cache.

Making it a project property means one could enable/disable it per project, which is not the desired behavior. Moreover, the script compilation build cache service is build scoped and doesn't have enough context to honor the flag on each invocation.

I think that using a system property instead would better capture the intent.

I confirmed that it can also be set in a `gradle.properties`  file, e.g.

```properties
systemProp.org.gradle.kotlin.dsl.caching.buildcache=true
```

and is then available when we need it to setup the script compilation cache.

@bamboo wdyt?

bamboo:
[...]

PR #1036, opened 2018-08-13
Let build cache integration be enabled via system property instead of project property

bamboo:
And prove it can be enabled via `$GRADLE_HOME/gradle.properties`.

See #1032

### DIFF
--- a/subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/caching/BuildCacheIntegrationTest.kt
+++ b/subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/caching/BuildCacheIntegrationTest.kt
@@ -59,7 +59,7 @@ class BuildCacheIntegrationTest : AbstractScriptCachingIntegrationTest() {
 
     @LeaksFileHandles("on the separate Gradle homes")
     @Test
-    fun `build cache integration is enabled via project property`() {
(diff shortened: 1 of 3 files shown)


subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/caching/BuildCacheIntegrationTest.kt
```diff
@@ -91,15 +91,22 @@ class BuildCacheIntegrationTest : AbstractScriptCachingIntegrationTest() {
...
-        // Cache hit from build cache
-        buildWithUniqueGradleHome("--build-cache", withBuildCacheIntegration).apply {
+        // Cache hit from build cache (enabled via gradle.properties file)
+        withUniqueGradleHome { gradleHome ->
 
-            compilationCache {
-                misses(cachedSettingsFile)
-                hits(cachedBuildFile)
-            }
+            File(gradleHome, "gradle.properties").writeText(
```
eskatos: :o: `gradleHome.resolve("gradle.properties")`?

bamboo: `gradleHome` is a `String`

PR #1037, opened 2018-08-13
WIP Introduce composite-build-logic sample

eskatos:
Still requires a work around for https://github.com/gradle/gradle-native/issues/522, see the `settings.gradle.kts` file.


JLLeitschuh:
@eskatos Can you also include in this sample how to setup the testing environment for the plugin as well?

If someone were to take this sample straight from the repository it would be awesome if the testing mechanisms were already in place.

samples/composite-build-logic/settings.gradle.kts
```diff
@@ -0,0 +1,20 @@
...
+// Work around for https://github.com/gradle/gradle-native/issues/522
+pluginManagement {
+    resolutionStrategy {
+        eachPlugin {
+            when (requested.id.id) {
+                "plugins.sample-plugin" -> useModule("$buildLogicGroup:plugins:latest-integration")
+            }
+        }
+    }
+}
```
JLLeitschuh: What?! Does this actually work!?

eskatos: Works fine from Gradle perspective. From my testing it seems IDEA can't import it though. I'll get back to this soon.

JLLeitschuh: Oh, that's really good information.

fuzzyweapon: Oh yeah I do this in my kotlin multiplatform (prior to their recent huge refactor of the MPP plugins) composite projects.  I haven't added native yet, so that might be while I don't have any issues with IDEA imports using this technique.

Issue #1021, opened 2018-08-09
Configuring Project Fails when non-jar in build classpath


jasongilanfarr:
### Expected Behavior
Configuring Project should succeed.

### Current Behavior
`WARN: error in opening zip file: /<hidden>/.gradle/caches/modules-2/files-2.1/<hidden>/2.0.22/30fb6082cb332d0d30f06a64814a26b4b5fda5e8/<hidden>.tar.gz
java.util.zip.ZipException: error in opening zip file
        at java.util.zip.ZipFile.open(Native Method)
        at java.util.zip.ZipFile.<init>(ZipFile.java:219)
        at java.util.zip.ZipFile.<init>(ZipFile.java:149)
        at java.util.zip.ZipFile.<init>(ZipFile.java:120)
        at org.jetbrains.kotlin.com.intellij.openapi.vfs.impl.ZipHandler$1.createAccessor(ZipHandler.java:45)
        at org.jetbrains.kotlin.com.intellij.openapi.vfs.impl.ZipHandler$1.createAccessor(ZipHandler.java:39)
        at org.jetbrains.kotlin.com.intellij.util.io.FileAccessorCache.createHandle(FileAccessorCache.java:60)
        at org.jetbrains.kotlin.com.intellij.util.io.FileAccessorCache.get(FileAccessorCache.java:52)
        at org.jetbrains.kotlin.com.intellij.openapi.vfs.impl.ZipHandler.getCachedZipFileHandle(ZipHandler.java:83)
        at org.jetbrains.kotlin.com.intellij.openapi.vfs.impl.ZipHandler.acquireZipHandle(ZipHandler.java:129)
        at org.jetbrains.kotlin.com.intellij.openapi.vfs.impl.ZipHandlerBase.createEntriesMap(ZipHandlerBase.java:44)
        at org.jetbrains.kotlin.com.intellij.openapi.vfs.impl.ArchiveHandler.getEntriesMap(ArchiveHandler.java:197)
        at org.jetbrains.kotlin.com.intellij.openapi.vfs.impl.jar.CoreJarHandler.<init>(CoreJarHandler.java:42)
        at org.jetbrains.kotlin.com.intellij.openapi.vfs.impl.jar.CoreJarFileSystem.lambda$new$0(CoreJarFileSystem.java:33)
        at org.jetbrains.kotlin.com.intellij.util.containers.ConcurrentFactoryMap$4.create(ConcurrentFactoryMap.java:224)
[...]

PR #1039, opened 2018-08-14
Let ClassBytesRepository ignore invalid classpath entries

eskatos:
In order to prevent unexpected failures when the classpath contains invalid entries that [should be ignored](https://docs.oracle.com/javase/8/docs/technotes/tools/findingclasses.html#userclass).

See #1021

### DIFF
--- a/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/ClassBytesRepository.kt
+++ b/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/ClassBytesRepository.kt
@@ -143,7 +143,7 @@ class ClassBytesRepository(classPath: ClassPath, classPathDependencies: ClassPat
  */
 private
 val File.isClassPathArchive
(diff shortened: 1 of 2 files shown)


subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/ClassBytesRepository.kt
```diff
@@ -138,6 +138,14 @@ class ClassBytesRepository(classPath: ClassPath, classPathDependencies: ClassPat
 }
 
 
+/**
+ * See https://docs.oracle.com/javase/8/docs/technotes/tools/findingclasses.html#userclass
+ */
+private
+val File.isClassPathArchive
+    get() = extension.equals("jar", ignoreCase = true) || extension.equals("zip", ignoreCase = true)
```
bamboo: :x: Compute `extension` only once with `extension.run { equals(...) || equals(...) }`

PR #1044, opened 2018-08-16
Report currently edited script location aware exceptions as IDE hints

eskatos:
This lets location aware runtime failures surface right in the editor at the offending line:

location-aware-hints

Script compilation failures are ignored as the native IDE hints already provides enough feedback. Only location aware exceptions for the currently edited script are reported. Reflection is used on the location aware exceptions because their java class comes from another class loader.


subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/KotlinBuildScriptDependenciesResolver.kt
```diff
@@ -221,3 +227,43 @@ object RequestQueue {
...
+        }
+    }
+    val cause = ex.cause ?: return null
+    return inferLineNumberAndMessageOrNull(scriptFile, cause)
+}
+
+
+private
+val Throwable.isReportedLocationAwareException
+    get() = this::class.java.name == LocationAwareException::class.java.name
```
bamboo: ❓ Why comparing the type name instead of checking the type via `is` or `as?`?

eskatos: Reflection is used on the location aware exceptions because their java class comes from another class loader. Using `is` or `as?` doesn't match.

PR #1060, opened 2018-08-23
Generate compilable accessors for extensions with types in the default package

### LINKED ISSUE #807: Generate compilable accessors for extension types in the default package
This could be fixed by adding import directives to the generated accessors source file, but only for those types in the default package to avoid name conflicts.

bamboo:
By adding the required import directives to the generated source files.

Resolves #807


eskatos:
👍

PR #1121, opened 2018-09-18
Test `build` on the Android sample in addition to existing smoke test

eskatos:
when ANDROID_HOME is set


subprojects/samples-tests/src/test/kotlin/org/gradle/kotlin/dsl/samples/samples.kt
```diff
@@ -34,3 +36,8 @@ fun withMergedGradleProperties(gradlePropertiesFile: File, action: () -> Unit) {
 private
 fun loadThenDeletePropertiesFrom(file: File) =
     loadPropertiesFrom(file).also { file.delete() }
+
+
+internal
+fun assumeAndroidHomeEnvVariableIsSet() =
```
bamboo: ⭕️ `assumeAndroidHomeIsSet` would be clear enough.

PR #1125, opened 2018-09-20
Upgrade to Kotlin 1.3

bamboo:
Upgrade Kotlin to the latest EAP release `1.3-rc-57` and automatically add the `kotlinEap()` repository to every project's `buildscript.repositories` and `repositories` to ease migration.


buildSrc/src/main/kotlin/plugins/kotlin-library.gradle.kts
```diff
@@ -27,18 +24,14 @@ plugins {
     id("org.gradle.kotlin.ktlint-convention")
 }
 
-kotlin {
-    experimental.coroutines = Coroutines.ENABLE
-}
-
```
eskatos: \o/

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/provider/KotlinScriptEvaluator.kt
```diff
@@ -277,6 +283,27 @@ class StandardKotlinScriptEvaluator(
 }
 
 
+private
+fun Settings.addKotlinEapRepository() {
+
+    gradle.settingsEvaluated {
+        if (pluginManagement.repositories.isEmpty()) {
```
eskatos: :x: the following would cover more cases:

```kotlin
pluginManagement.repositories.run {
    val wasEmpty = isEmpty()
    kotlinEap()
    if (wasEmpty) {
        gradlePluginPortal()
    }
}
```

with the remaining caveat that a build that already has kotlin-eap will have it twice, in line with what is done for project and their buildscript below.

bamboo:
Thanks, @eskatos!

I've created #1131 to track the issue related to the use of`ProjectBuilder` and the kapt issue went away after the sample and wrappers were all upgraded to `1.3.0-rc-57`

PR #1129, opened 2018-09-21
Fix usabililty of Configuration accessors in the `configurations {}` block

eskatos:
by generating accessors to the existing container elements, leveraging `NamedDomainObjectCollection.collectionSchema`.

Those extensions expose the internal `DefaultConfiguration` type because of https://github.com/gradle/gradle-native/issues/856

Old accessors removed as they conflict with the new and work in less situations. Extensions were added to provide source-compatibility to scripts that were using the old accessors, marked as deprecated and with IDE guidance for replacement.

With the example from the original report in #1118 this is now possible:
```kotlin
configurations {
    implementation {
        exclude(group = "org.foo")
    }
}
```

PR #1130, opened 2018-09-21
Accessors for existing elements of `tasks` and containers

### LINKED ISSUE #1118: Support statically-resolved configurations in the configurations block
Despite applying the Kotlin plugin declaratively in the `plugins {}` block:

```
plugins {
	kotlin("jvm") version "1.2.70"
}
```

...I can only resolve configurations in the `configurations {}` block dynamically:

```
configurations {
	getByName("implementation") {
		exclude(group = "org.foo")
	}
}
```

The [migration guide](https://guides.gradle.org/migrating-build-logic-from-groovy-to-kotlin/#configurations-and-dependencies) says:

> Each configuration contributed by an applied plugin is also available as a member of the configurations container, so you can reference it just like any other configuration.
>
> *Knowing what configurations are available*
> The easiest way to find out what configurations are available is by asking your IDE for suggestions within the configurations container.
>
> You can also use the `:kotlinDslAccessorsReport` task, which prints the Kotlin code for accessing the configurations contributed by applied plugins and provides the names for all of those accessors.

Indeed the `:kotlinDslAccessorsReport` task prints:

```
/**
 * The 'implementation' configuration.
 */
val ConfigurationContainer.`implementation`: Configuration
    get() = getByName("implementation")
```

I see with Quick Documentation in IntelliJ that `dependencies` has a Kotlin DSL `Project` extension function:

> org.gradle.kotlin.dsl ProjectExtensionsKt.class
> public fun Project.dependencies(
>    configuration: DependencyHandlerScope.() → Unit
>): Unit

...whereas `configurations` does not:

> org.gradle.api.Project
> public abstract org.gradle.api.artifacts.ConfigurationContainer getConfigurations()

It would be great if the `implementation`, etc. extension properties were statically available in `configurations {}` as they are in `dependencies {}`):

```
configurations {
	implementation {
		exclude(group = "org.foo")
	}
}
```

eskatos:
by leveraging the mechanics introduced by #1129 to fix #1118.

### DIFF
--- a/subprojects/integ-tests/src/test/kotlin/org/gradle/kotlin/dsl/integration/ProjectSchemaAccessorsIntegrationTest.kt
+++ b/subprojects/integ-tests/src/test/kotlin/org/gradle/kotlin/dsl/integration/ProjectSchemaAccessorsIntegrationTest.kt
@@ -853,6 +853,63 @@ class ProjectSchemaAccessorsIntegrationTest : AbstractPluginIntegrationTest() {
         build("help")
     }
 
+    @Test
+    fun `accessors to existing source sets`() {
+
+        withBuildScript("""
+            plugins {
+                java
+            }
+
+            sourceSets {
+                main {
+                    java {
+                        srcDir("src/main/java-too")
+                    }
+                }
+                val integTest by registering {
+                    java.srcDir(file("src/integTest/java"))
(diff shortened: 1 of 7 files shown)


gildor:
Even for tasks? it means that you have access to tasks configuration with type-safe accessor?

PR #1133, opened 2018-09-23
Refine messages reported to the IDE editor and their lifecycle

eskatos:
This PR lets text of IDE reports be simpler and actionable and also refine how failures are reported, distinguishing those happening in the currently edited script from those coming from other parts of the build logic. See the added coverage for handled cases.

To try it out:
```
./gradlew customInstallation
```
Then import a build in IntelliJ using the Gradle installation from `./build/custom`

This PR also introduces support for location aware hints for build configuration runtime failures originating from the currently edited script. This feature is disabled by default and can be enabled using an internal property. There are some UX issues to iron out but this will allow us to try it out, see the added coverage.

This PR is part of #89


eskatos:
@oehme, this should address some of your feedback

PR #1141, opened 2018-09-25
Remove support for deferred configurable extensions

marcphilipp:
### Context

Gradle 5.0 will remove `@DeferredConfigurable`.

See gradle/gradle#6276

### Contributor Checklist
- [x] Make sure that all commits are [signed off](https://git-scm.com/docs/git-commit#git-commit---signoff) to indicate that you agree to the terms of [Developer Certificate of Origin](https://developercertificate.org/).

### DIFF
--- a/subprojects/plugins/src/test/kotlin/org/gradle/kotlin/dsl/plugins/precompiled/PrecompiledScriptPluginTest.kt
+++ b/subprojects/plugins/src/test/kotlin/org/gradle/kotlin/dsl/plugins/precompiled/PrecompiledScriptPluginTest.kt
@@ -259,6 +259,8 @@ class PrecompiledScriptPluginTest : AbstractPluginTest() {
 
                     $pluginManagementBlock
 
+                    enableFeaturePreview("STABLE_PUBLISHING")
+
                 """)
 
                 withFile("build.gradle.kts", """
--- a/samples/gradle-plugin/plugin/settings.gradle.kts
+++ b/samples/gradle-plugin/plugin/settings.gradle.kts
@@ -1,0 +1,1 @@
+enableFeaturePreview("STABLE_PUBLISHING")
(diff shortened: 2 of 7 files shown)


subprojects/plugins/src/test/kotlin/org/gradle/kotlin/dsl/plugins/precompiled/PrecompiledScriptPluginTest.kt
```diff
@@ -259,6 +259,8 @@ class PrecompiledScriptPluginTest : AbstractPluginTest() {
 
                     $pluginManagementBlock
 
+                    enableFeaturePreview("STABLE_PUBLISHING")
```
bamboo: ❓ does the test fail without this?

marcphilipp: Yes, it does because in the currently used Gradle snapshot (as in `master`) the publishing block still uses `@DeferredConfigurable` by default. The code I changed causes the block to be evaluated which then fails these builds with a "cannot change publishing block after it has been evaluated" message.

bamboo: Ok. So if I understand it correctly, we will be able to remove those feature flags once we are on a new wrapper built on top of latest  Gradle master, right?

marcphilipp: Yep, once the change is merge there. To do so, we need this change merged here first, though. Cyclic dependencies are so much fun... 😉

bamboo: Indeed :)

bamboo:
Thanks, @marcphilipp!

Issue #1131, opened 2018-09-21
`FastStringServiceFactory` class not found exception in `ProjectBuilder` based tests after upgrade to latest Gradle master


bamboo:
`TaskContainerDslTest` and `NamedContainerDslTest` started failing after upgrading to a distribution built from latest Gradle master with:
```
java.lang.NoClassDefFoundError: org/gradle/internal/impldep/org/apache/groovy/json/FastStringServiceFactory
	at org.gradle.initialization.DefaultGradleApiSpecProvider.getExportedClasses(DefaultGradleApiSpecProvider.java:29)
	at org.gradle.initialization.GradleApiSpecAggregator.mergeSpecsOf(GradleApiSpecAggregator.java:63)
	at org.gradle.initialization.GradleApiSpecAggregator.aggregate(GradleApiSpecAggregator.java:45)
	at org.gradle.initialization.DefaultClassLoaderRegistry.apiSpecFor(DefaultClassLoaderRegistry.java:44)
	at org.gradle.initialization.DefaultClassLoaderRegistry.restrictToGradleApi(DefaultClassLoaderRegistry.java:35)
	at org.gradle.initialization.DefaultClassLoaderRegistry.<init>(DefaultClassLoaderRegistry.java:29)
	at org.gradle.internal.service.scopes.GlobalScopeServices.createClassLoaderRegistry(GlobalScopeServices.java:216)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:73)
	at org.gradle.internal.service.ReflectionBasedServiceMethod.invoke(ReflectionBasedServiceMethod.java:35)
	at org.gradle.internal.service.DefaultServiceRegistry$FactoryMethodService.invokeMethod(DefaultServiceRegistry.java:821)
	at org.gradle.internal.service.DefaultServiceRegistry$FactoryService.create(DefaultServiceRegistry.java:774)
	at org.gradle.internal.service.DefaultServiceRegistry$ManagedObjectServiceProvider.getInstance(DefaultServiceRegistry.java:568)
[...]

PR #1142, opened 2018-09-25
Let this build support ProjectBuilder based tests after the upgrade to Groovy 2.5

eskatos:
See #1131


eskatos:
This needs more work

PR #1146, opened 2018-09-27
Remove usage of StartParameter.isRecompileScripts

marcphilipp:
The `--recompile-scripts` option will be removed from Gradle 5.0.

See gradle/gradle#6306


bamboo:
The PR should be targeting develop.

PR #1148, opened 2018-09-27
Dogfood recent improvements in samples

eskatos:


JLLeitschuh:
Are all these tasks configured lazily using this new API or eagerly?
Also, does the behaviour difference between groovy and Kotlin?

```
tasks {
    test {
          // Is this block executed eagerly in groovy by lazily in Kotlin?
    }
}
```

PR #1152, opened 2018-09-28
Let build cache integration be disabled via system property

bamboo:
Before this commit, build cache integration had to be explicitly enabled for Kotlin DSL scripts by setting the value of the system property `org.gradle.kotlin.dsl.caching.buildcache` to `true`.

With this commit, build cache integration is enabled by default and can be disabled by setting the value of said system property to`false`.


JLLeitschuh:
Is this feature documented in the Gradle docs somewhere?

PR #1156, opened 2018-10-02
ntroduce type-safe accessors for plugin ids 

bamboo:
With this PR the Kotlin DSL will generate type-safe accessors for all plugin ids found in the `buildSrc` classpath.

So instead of writing:

```kotlin
plugins {
    id("gradlebuild.strict-compile")
    id("gradlebuild.classcycle")
}
```

One can write:

```kotlin
plugins {
    gradlebuild.`strict-compile`
    gradlebuild.classcycle
}
```

All while enjoying content assist and source code navigation all the way down to the plugin implementation class.

### DIFF
--- a/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/accessors/PluginAccessorsClassPath.kt
+++ b/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/accessors/PluginAccessorsClassPath.kt
@@ -1,0 +1,200 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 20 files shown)


subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/accessors/PluginAccessorsClassPath.kt
```diff
@@ -0,0 +1,200 @@
...
+fun pluginDependenciesSpecOf(extendedType: String): String = when (extendedType) {
+    "PluginDependenciesSpec" -> "this"
+    else -> "plugins"
+}
+
+
+private
+inline fun <T> Sequence<T>.runEach(f: T.() -> Unit) {
+    forEach { it.run(f) }
+}
```
JLLeitschuh: Nice.

PR #1164, opened 2018-10-03
 Generate compilable accessors for container elements of default package type

bamboo:


subprojects/integ-tests/src/test/kotlin/org/gradle/kotlin/dsl/integration/ProjectSchemaAccessorsIntegrationTest.kt
```diff
@@ -129,41 +117,83 @@ class ProjectSchemaAccessorsIntegrationTest : AbstractPluginIntegrationTest() {
...
-                    withFile("plugin.gradle.kts", """
-                        tasks.register<CustomTask>("customTask")
-                    """)
-                }
+                    val beans = DefaultPolymorphicDomainObjectContainer(
+                        Any::class.java,
+                        DirectInstantiator.INSTANCE
+                    ) { (it as Bean).name }
+
+                    // TODO: schema doesn't expose the registered element as Bean
```
eskatos: This looks like a bug to me, polymorphic containers are supposed to expose the implementation type in their schema.

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/DefaultPackageTypesTest.kt
```diff
@@ -32,4 +34,29 @@ class DefaultPackageTypesTest {
             equalTo(listOf("Extension"))
         )
     }
+
+    @Test
+    fun `#importsRequiredBy takes container elements into account`() {
```
JLLeitschuh: ❓  What's the significance of the `#` symbol in the test name?

bamboo: That's meant to signify the name refers to a function (like in javadoc links).

JLLeitschuh: Ah, thanks.

subprojects/integ-tests/src/test/kotlin/org/gradle/kotlin/dsl/integration/ProjectSchemaAccessorsIntegrationTest.kt
eskatos: :x: This won't affect public api consumers as there's no public way to create polymorphic containers. This test should be removed.

PR #1174, opened 2018-10-11
Let IDE resolver not run buildSrc tests for both leniency and better responsiveness

eskatos:


melix:
Yes, please! :tada:

subprojects/tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/BuildSrcClassPathModeConfigurationAction.kt
```diff
@@ -38,16 +39,26 @@ import org.gradle.kotlin.dsl.resolver.buildSrcSourceRootsFilePath
...
+                disableAllTestTasks()
                 configureBuildSrcSourceRootsTask()
             }
         }
     }
 
+    private
+    fun Project.disableAllTestTasks() {
+        allprojects {
+            it.tasks.withType<Test> {
```
bamboo: ❌ `.withType<Test>().configureEach {`?

subprojects/tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/BuildSrcClassPathModeConfigurationAction.kt
```diff
@@ -38,16 +39,26 @@ import org.gradle.kotlin.dsl.resolver.buildSrcSourceRootsFilePath
 
 
 internal
-class BuildSrcSourceRootsConfigurationAction : BuildSrcProjectConfigurationAction {
+class BuildSrcClassPathModeConfigurationAction : BuildSrcProjectConfigurationAction {
```
bamboo: 👍

JLLeitschuh:
Shouldn't this be implemented in the Gradle parent repo not here?
What about tests that are written in Groovy or Java?

JLLeitschuh:
Okay.
This makes sense. Thanks.

> What could be done in gradle/gradle is completely remove buildSrc tests from configuration time, this would make this change irrelevant but is another story.

So this change doesn't break having the tests run when a user runs `./gradlew` from the parent project directory then. The `buildSrc` tests time becomes pretty much irrelevant when you have the build cache enabled.

eskatos:
@JLLeitschuh this changes apply only when fetching script dependencies from the IDE editor, it doesn't change anything else

PR #1181, opened 2018-10-16
Add KotlinClosure3 for interop with trivariate Closures

eskatos:


subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/GroovyInteroperability.kt
```diff
@@ -121,6 +121,30 @@ class KotlinClosure2<in T : Any?, in U : Any?, V : Any>(
 }
 
 
+/**
+ * Adapts a trivariate Kotlin function to a binary Groovy [Closure].
```
bamboo: :o: I would use _ternary_ here instead of _trivariate_ since we are using `unary` and `binary` for the other types.

bamboo: ❓You mean `ternary` Groovy closure, right?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/GroovyInteroperability.kt
```diff
@@ -121,6 +121,30 @@ class KotlinClosure2<in T : Any?, in U : Any?, V : Any>(
...
+ * @param owner optional owner of the Closure.
+ * @param thisObject optional _this Object_ of the Closure.
+ *
+ * @see [Closure]
+ */
+class KotlinClosure3<in T : Any?, in U : Any?, in V : Any?, R : Any>(
+    val function: (T, U, V) -> R?,
+    owner: Any? = null,
+    thisObject: Any? = null
+) : Closure<R?>(owner, thisObject) {
```
JLLeitschuh: Do you want to expose constructors for these directly instead of the `of` methods like you were doing elsewhere?

eskatos: Yes

PR #1188, opened 2018-10-17
Precompute message in order not to leak Project instance

eskatos:
The KotlinCompile task logger goes through RMI to the Kotlin Compiler daemon. Holding a Project instance per task exacerbated a memory leak in the Kotlin compiler infrastructure.

This only impacted `KotlinCompile` tasks with the `kotlin-dsl` plugin applied.

Also see https://youtrack.jetbrains.com/issue/KT-27639


subprojects/plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslCompilerPlugins.kt
```diff
@@ -93,11 +93,11 @@ fun KotlinCompile.replaceLoggerWith(logger: Logger) {
 private
 class KotlinCompilerWarningSubstitutingLogger(
     private val delegate: Logger,
-    private val project: Project
+    private val kotlinDslPluginExperimentalWarning: String
```
oehme: Could be the project name so you don't have to preallocate the whole message.

PR #1192, opened 2018-10-18
Add implicit import for java.util.concurrent.Callable

eskatos:
given it's required for many Gradle API when deferring computation, e.g. `copySpec { from(Callable { .. }) }`

Note that `copySpec { from(provider { .. }) }` also works


JLLeitschuh:
Oooh. Excellent idea!

oehme:
Shouldn't we use the Provider API to defer computations?

JLLeitschuh:
@oehme You can't create a `Provider` without using a `Callable` currently. Also, `Provider` isn't a functional interface.

PR #1195, opened 2018-10-18
Remove support for discontinued json project schema snapthots

eskatos:
Warn and suggest to delete the now unnecessary file.


subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/KotlinDslAccessorsSnapshotTaskIntegrationTest.kt
```diff
@@ -1,299 +0,0 @@
-package org.gradle.kotlin.dsl.accessors
```
JLLeitschuh: I'd say it's valuable to keep this test around in some incarnation to ensure that the `ProjectSchema` object is at least being generated correctly.

If you can find an alternative way to expose the `ProjectSchema` in some helpful way, this test still contains value.

eskatos: Right! Adapted and re-added

bamboo: I'm sorry for being the party pooper but that test was getting in my way.

JLLeitschuh: That's unfortunate, oh well.
Care to explain how it was getting in your way? Asking out of curiosity.

PR #1199, opened 2018-10-22
Reduce time required to generate the project accessors

bamboo:
By emitting the relevant bytecode directly instead of calling the Kotlin compiler to do it.


subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/AccessorBytecodeEmitterSpike.kt
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.Test
+
+
+inline fun <T : Dependency> DependencyHandler.foo(
+    dependency: T,
+    action: Action<T>
+): T = dependency.also { action.execute(it) }
+
+
+class AccessorBytecodeEmitterSpike : TestWithTempFiles() {
```
eskatos: :x: rename to eg. `KotlinMetadataTest`?

eskatos:
@JLLeitschuh, I agree this makes contributing new accessors a bit harder. But I think it's not as bad as it sounds. A proposal for new accessors should start with some example source we all can copy and paste in our build scripts and try. The next step is to turn that into a source template we can generate, with its inputs. The only thing this PR changes is adding the bytecode generation.

It shouldn't be a lot of bytecode anyway, just like the source for accessors shouldn't be a lot of code. They should be lean, the complex things being done by the Gradle model. And there shouldn't be many of them.

DSL extensibility and collections elements are good because we can reuse them a lot, uniformly. For example, your PR to add accessors to extensions of `DependencyHandler` will not need any change to or addition of bytecode generation code.

We should also make the code simpler to work with. There's room for improvement, see my review comment above for some first steps.

PR #1240, opened 2018-11-08
#390 Document how to leverage the allopen compiler plugin

jmfayard:
### Context

If you don't declare your Task, Project, Extensions, ... as an `open class `,
Gradle 5.0 will fails at runtime with the error

  `> Cannot create a proxy class for final class at runtime.`

This will be tentatively fixed in Gradle 5.1 by applying automatically the "allopen" compiler plugin.

I thought it could be useful in the meantime to have the work-around documented as code in one of the samples. 

See discussion here https://github.com/gradle/kotlin-dsl/issues/390

### Contributor Checklist
- [x] Make sure that all commits are [signed off](https://git-scm.com/docs/git-commit#git-commit---signoff) to indicate that you agree to the terms of [Developer Certificate of Origin](https://developercertificate.org/).

### DIFF
--- a/samples/project-with-buildSrc/buildSrc/build.gradle.kts
+++ b/samples/project-with-buildSrc/buildSrc/build.gradle.kts
@@ -1,11 +1,9 @@
 plugins {
     `kotlin-dsl`
-
-    // see buildSrc/src/main/kotlin/allopen-compiler-plugin.kt
-    id("org.jetbrains.kotlin.plugin.allopen") version "1.3.0"
+    kotlin("plugin.allopen") version KotlinVersion.CURRENT.toString()
 }
 
-// see buildSrc/src/main/kotlin/allopen-compiler-plugin.kt
+// see buildSrc/src/main/kotlin/my/plugin.allopen.kt
 allOpen {
     annotation("my.AllOpen")
 }
--- a/samples/project-with-buildSrc/buildSrc/build.gradle.kts
+++ b/samples/project-with-buildSrc/buildSrc/build.gradle.kts
@@ -1,6 +1,15 @@
 plugins {
     `kotlin-dsl`
+
+    // see buildSrc/src/main/kotlin/allopen-compiler-plugin.kt
+    id("org.jetbrains.kotlin.plugin.allopen") version "1.3.0"
 }
+
+// see buildSrc/src/main/kotlin/allopen-compiler-plugin.kt
+allOpen {
+    annotation("my.AllOpen")
+}
+
 
 repositories {
     jcenter()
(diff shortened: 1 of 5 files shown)


samples/project-with-buildSrc/buildSrc/build.gradle.kts
```diff
@@ -1,7 +1,16 @@
 plugins {
     `kotlin-dsl`
+
+    // see buildSrc/src/main/kotlin/allopen-compiler-plugin.kt
+    id("org.jetbrains.kotlin.plugin.allopen") version "1.3.0"
```
JLLeitschuh:
```suggestion
    id("org.jetbrains.kotlin.plugin.allopen") version KotlinVersion.CURRENT.toString()
```

I know this is different from what the other plugins use, but it will keep the plugin version in sync with the version of Kotlin that the Kotlin DSL is using.

samples/project-with-buildSrc/buildSrc/src/main/kotlin/MyProjectExtension.kt
```diff
@@ -17,8 +17,8 @@ import org.gradle.api.model.ObjectFactory
 
 import org.gradle.kotlin.dsl.*
 
-
-open class MyProjectExtension(objects: ObjectFactory) {
+@my.AllOpen
+class MyProjectExtension(objects: ObjectFactory) {
```
JLLeitschuh: Is this really any better?

samples/project-with-buildSrc/buildSrc/build.gradle.kts
jmfayard: Ok! Should I use the `kotlin("allopen")` syntax?

samples/project-with-buildSrc/buildSrc/src/main/kotlin/MyProjectExtension.kt
jmfayard: Not really, I would say both are bad!
The problem is that there are not common interface for Project extensions AFAIK. 

A solution would be to add to an empty interface 

```kt
// MyProjectExtension.kt
class MyProjectExtension(objects: ObjectFactory) : my.Extension

// my/allopen-compiler-plugin.kt
package my

@AllOpen 
interface Extension
```

samples/project-with-buildSrc/buildSrc/build.gradle.kts
JLLeitschuh: Not sure. Sorry.

jmfayard: Updated to `kotlin("plugin.allopen") version KotlinVersion.CURRENT.toString()`

samples/project-with-buildSrc/buildSrc/src/main/kotlin/MyProjectExtension.kt
jmfayard: I went forward and did that

PR #1246, opened 2018-11-14
 Let `creating` property delegate use `create` instead of `register`

bamboo:
To minimize confusion when migrating away from the eager API.


JLLeitschuh:
The config avoidance API is what the new API Gradle wants users to be using moving forward.
These changes seem to be at odds with that goal.  I'm confused.

eskatos:
@JLLeitschuh use `registering` or `existing` for configuration avoidance. `creating` and `getting` should be eager, they should simply delegate to `create()` and `getByName()`, just like `registering` and `existing` respectively delegate to `register()` and `named()`. This change and #1247 makes the eager delegated properties behavior more coherent, less surprising.

JLLeitschuh:
Okay, this makes sense. I'm glad that there's an alternative that still provides support for config avoidance.

PR #1247, opened 2018-11-14
Let `getting` property delegate use `getByName` instead of `named`

bamboo:
To minimize confusion when migrating away from the eager API.


JLLeitschuh:
Have there been reports of users finding this to be confusing as it was before?

eskatos:
@JLLeitschuh yes, it was surprising and caused confusion in knowing what is eager and what is lazy.

PR #1248, opened 2018-11-14
Kotlin 1.3.10

bamboo:
https://github.com/JetBrains/kotlin/releases/tag/v1.3.10


samples/hello-coroutines/build.gradle.kts
```diff
@@ -2,7 +2,7 @@ import org.jetbrains.kotlin.gradle.dsl.Coroutines
 
 plugins {
     application
-    kotlin("jvm") version "1.3.0"
+    kotlin("jvm") version "1.3.10"
```
JLLeitschuh: Consider this instead?

```suggestion
    kotlin("jvm") version KotlinVersion.CURRENT.toString()
```

bamboo: Thanks for the suggestion. We want the samples to advertise the version explicitly.

PR #1258, opened 2018-11-23
Migrate to the latest Gradle Build Cache API

bamboo:
- Replace `OriginTaskExecutionMetadata` by `OriginMetadata`

### DIFF
--- a/subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/caching/LoadDirectoryTest.kt
+++ b/subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/caching/LoadDirectoryTest.kt
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.kotlin.dsl.caching
+
+import org.gradle.internal.id.UniqueId
+
+import org.gradle.kotlin.dsl.cache.LoadDirectory
+import org.gradle.kotlin.dsl.cache.PackMetadata
+import org.gradle.kotlin.dsl.cache.ScriptBuildCacheKey
+
+import org.gradle.kotlin.dsl.fixtures.TestWithTempFiles
+
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+import java.io.InputStream
+
+
+class LoadDirectoryTest : TestWithTempFiles() {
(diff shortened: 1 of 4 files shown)


wolfs:
@bamboo: Why do you think that using `RootBuildCacheControllerRef` is a problem? The best way would probably be to use the `Gradle` scoped `BuildCacheController`.

For nested builds, the `RootBuildCacheControllerRef` would already be available before the `Gradle` object, so you could even cache settings scripts from there, right? I don't know if it is worth the effort though.

bamboo:
@wolfs thanks for taking a look.

I was thinking the use might be problematic because of the call to `forNonRootBuild`, it's always protected by a call to `isSet` though so should be fine?

`ScriptCache` is a build service so it can't depend on `BuildCacheController` directly.

wolfs:
@bamboo Yeah, it is protected, though it throws an exception. Do we always expect that the `RootBuildCacheControllerRef` is set when you query it? The I would just call `forNonRootBuild`, else maybe we should check `isSet` first.

bamboo:
Thinking a litte more about this, having the Kotlin DSL call `forNonRootBuild` seems to imply more knowledge than what the Kotlin DSL currently has (is it really a non root build? does it matter?) but the tests pass and looking at the code it seems harmless so I think the PR is good to go as is.

wolfs:
Probably the method shouldn't be called `forNonRootBuild` but something like `getExisting()`.

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/caching/LoadDirectoryTest.kt
```diff
@@ -0,0 +1,60 @@
...
+
+        // then:
+        assert(
+            subject.load(packProducedBy(currentBuildInvocation))!!.metadata.isProducedByCurrentBuild
+        )
+
+        // and:
+        assert(
+            !subject.load(packProducedBy(previousBuildInvocation))!!.metadata.isProducedByCurrentBuild
+        )
```
JLLeitschuh: Why not use `assertFalse`?

lptr:
I don't think we should do this. Instead, Kotlin DSL should migrate to the new execution model we are going to use for cacheable work. I wouldn't start using that API in an external project, though, as we don't yet want to guarantee its stability.

Instead we should get Kotlin DSL merged into `gradle/gradle`. Once that's done, the build cache team can take care of migration to the new API.

bamboo:
@lptr, thanks for the input.

How about keeping the PR only about the migration from the deprecated `OriginTaskExecutionMetadata` to `OriginMetadata`?

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/caching/LoadDirectoryTest.kt
bamboo: For symmetry.

JLLeitschuh: 😕 Symmetry? Aren't `assertFalse` and `assertTrue` pretty standard?

bamboo: The code is not using `assertTrue`.

bamboo: Moved to `assertFalse` / `assertTrue`.

Issue #1243, opened 2018-11-08
Task `generateScriptPluginAdapters` from the `kotlin-dsl` plugin should maximise cacheability


bamboo:
## Expected Behavior

The output of `generateScriptPluginAdapters` task can be cached and retrieved from the cache no matter where the project is located in the file system.

## Current Behavior 

Due to how [inputs are captured](https://github.com/gradle/kotlin-dsl/blob/6300ca465fd18af2941d48978c9586fdd1aedfe5/subprojects/plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/precompiled/PrecompiledScriptPlugins.kt#L153), caching is sensitive to the location of the project in the file system:

![48189487-a1fe2680-e340-11e8-9fb7-7c15fc43afc2](https://user-images.githubusercontent.com/51689/48215797-287b2e00-e36a-11e8-92b2-4797a3e23f96.png)

PR #1275, opened 2018-12-05
Let GenerateScriptPluginAdapters task be cacheable

eskatos:
See #1243

### DIFF
--- a/subprojects/plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/precompiled/GenerateScriptPluginAdapters.kt
+++ b/subprojects/plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/precompiled/GenerateScriptPluginAdapters.kt
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.gradle.kotlin.dsl.plugins.precompiled
+
+import org.gradle.api.DefaultTask
+import org.gradle.api.file.FileTree
+import org.gradle.api.tasks.CacheableTask
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.Internal
+import org.gradle.api.tasks.PathSensitive
+import org.gradle.api.tasks.PathSensitivity
+import org.gradle.api.tasks.OutputDirectory
+import org.gradle.api.tasks.TaskAction
+
+import java.io.File
+
+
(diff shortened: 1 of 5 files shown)


subprojects/plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/precompiled/GenerateScriptPluginAdapters.kt
```diff
@@ -0,0 +1,104 @@
...
+    outputFile.writeText("""
+
+        $packageDeclaration
+
+        /**
+         * Precompiled [$scriptFileName][$compiledScriptTypeName] script plugin.
+         *
+         * @see $compiledScriptTypeName
+         */
+        class $simplePluginAdapterClassName : org.gradle.api.Plugin<$targetType> {
```
JLLeitschuh: I can't come up with a concrete example of where this might blow up, but I think to be safe, the name of the class needs to be wrapped in backticks.
```
class `$simplePluginAdapterClassName` : org.gradle.api.Plugin<$targetType> {
```

bamboo: That's not necessary for `simplePluginAdapterClassName` is guaranteed to be a valid Java identifier per https://github.com/gradle/kotlin-dsl/blob/071d76cf5c78736f00c22fd6240ed425df3acf51/subprojects/plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/precompiled/ScriptPlugin.kt#L59-L63

That assumption could be made clearer with an inline class though.

JLLeitschuh: Okay, I think I have an example.
If your script is named, `🦆.gradle.kts`.
A class called `🦆Plugin` will not compile without the backticks.
I'm guessing the same problem exists if you have spaces in the name of your file like `a great name.gradle.kts`.

eskatos: @JLLeitschuh, this should respectively be `_Plugin` and `Great_namePlugin`, have another look at Rodrigo's comment above and see what `asJavaIdentifier()` does.

JLLeitschuh: Right, okay. I understand.
A quick test that demonstrates the exact problem you're referring to:
```kotlin
@Test
fun `script name collision test`() {
    val script1 =
        newFile("file#22.gradle.kts")
    val script2 =
        newFile("file,22.gradle.kts")

    val scriptPlugin1 = ScriptPlugin(script1)
    val scriptPlugin2 = ScriptPlugin(script2)

    assertThat(
        scriptPlugin1.simplePluginAdapterClassName,
        not(equalTo(scriptPlugin2.simplePluginAdapterClassName)))
}
```
This currently fails because the two class names are the same.

JLLeitschuh: @eskatos Followup here: https://github.com/gradle/kotlin-dsl/issues/1284

bamboo:
Required one more change done in 83e3686f2e7feb928beca9e98200f0203ba86662.

PR #1278, opened 2018-12-05
Generate type-safe accessors for extensions of project.dependencies

eskatos:
This PR adds generation of type-safe accessors for extensions of `project.dependencies`.

Note that `DependencyHandler` doesn't expose that it implements `ExtensionAware`, this PR doesn't change that and does not expose it either on `DependencyHandlerScope`.

See #1185

### DIFF
--- a/subprojects/integ-tests/src/test/kotlin/org/gradle/kotlin/dsl/integration/ProjectSchemaAccessorsIntegrationTest.kt
+++ b/subprojects/integ-tests/src/test/kotlin/org/gradle/kotlin/dsl/integration/ProjectSchemaAccessorsIntegrationTest.kt
@@ -1000,6 +1000,35 @@ class ProjectSchemaAccessorsIntegrationTest : AbstractPluginIntegrationTest() {
         build("help")
     }
 
+    @Test
+    fun `accessors to extensions of the dependency handler`() {
+
+        withKotlinBuildSrc()
+        withFile("buildSrc/src/main/kotlin/Mine.kt", """
+            open class Mine {
+                val some = 19
+                val more = 23
+            }
+        """)
+        withFile("buildSrc/src/main/kotlin/my-plugin.gradle.kts", """
+            (dependencies as ExtensionAware).extensions.create<Mine>("mine")
+        """)
+
+        withBuildScript("""
+            plugins {
+                `my-plugin`
+            }
+
+            dependencies {
+                println(mine.some + project.dependencies.mine.more)
+            }
+        """.trimIndent())
+
+        build("help").apply {
+            assertThat(output, containsString("42"))
+        }
+    }
+
     private
     fun withBuildSrc(contents: FoldersDslExpression) {
         withFolders {
--- a/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/DependencyHandlerScope.kt
+++ b/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/DependencyHandlerScope.kt
@@ -21,6 +21,7 @@ import org.gradle.api.artifacts.Dependency
 import org.gradle.api.artifacts.ExternalModuleDependency
 import org.gradle.api.artifacts.ModuleDependency
 import org.gradle.api.artifacts.dsl.DependencyHandler
+import org.gradle.api.plugins.ExtensionAware
 
 
 /**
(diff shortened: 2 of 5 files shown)


subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/DependencyHandlerScope.kt
```diff
@@ -21,21 +21,27 @@ import org.gradle.api.artifacts.Dependency
...
 
     companion object {
         fun of(dependencies: DependencyHandler) =
-            DependencyHandlerScope(dependencies)
+            ExtensionAwareDependencyHandlerScope(dependencies) as DependencyHandlerScope
+
+        private
+        class ExtensionAwareDependencyHandlerScope(
+            dependencies: DependencyHandler
+        ) : DependencyHandlerScope(dependencies), ExtensionAware by (dependencies as ExtensionAware)
```
JLLeitschuh: Interesting way of handling this. Can you elaborate as to why you chose to go this route instead of adding `ExtensionAware` to `DependencyHandlerScope`?
By doing this, users in their build scripts can't access the rest of the methods that come along with having a type be `ExtensionAware`. For example:
```kotlin
dependencies {
    val someStoredString: String by this
}
```

Just trying to understand the thought process here.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/DependencyHandlerScope.kt
```diff
@@ -21,21 +21,27 @@ import org.gradle.api.artifacts.Dependency
...
-class DependencyHandlerScope
+open class DependencyHandlerScope
 private constructor(
     val dependencies: DependencyHandler
 ) : DependencyHandler by dependencies {
 
     companion object {
         fun of(dependencies: DependencyHandler) =
-            DependencyHandlerScope(dependencies)
+            ExtensionAwareDependencyHandlerScope(dependencies) as DependencyHandlerScope
```
JLLeitschuh: Instead of casting (which then I think needs to happen at runtime as well (but I might be wrong)).
Just explicitly declare the return type.
```kotlin
fun of(dependencies: DependencyHandler): DependencyHandlerScope
```

subprojects/integ-tests/src/test/kotlin/org/gradle/kotlin/dsl/integration/ProjectSchemaAccessorsIntegrationTest.kt
```diff
@@ -1000,6 +1000,35 @@ class ProjectSchemaAccessorsIntegrationTest : AbstractPluginIntegrationTest() {
...
+                `my-plugin`
+            }
+
+            dependencies {
+                println(mine.some + project.dependencies.mine.more)
+            }
+        """.trimIndent())
+
+        build("help").apply {
+            assertThat(output, containsString("42"))
```
JLLeitschuh: I love the use of 42 everywhere. 😆

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/DependencyHandlerScope.kt
eskatos: Adding `ExtensionAware` to `DependencyHandlerScope` would bring non-uniformity and confusion because it would not be available on `DependencyHandler. Taking one step at a time.

JLLeitschuh: Fair enough.

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/DependencyHandlerExtensionsTest.kt
```diff
@@ -309,4 +311,8 @@ class DependencyHandlerExtensionsTest {
             verifyNoMoreInteractions()
         }
     }
+
+    private
+    fun newDependencyHandlerMock(stubbing: KStubbing<DependencyHandler>.(DependencyHandler) -> Unit = {}) =
```
bamboo: 👍

subprojects/integ-tests/src/test/kotlin/org/gradle/kotlin/dsl/integration/ProjectSchemaAccessorsIntegrationTest.kt
```diff
@@ -1000,6 +1000,35 @@ class ProjectSchemaAccessorsIntegrationTest : AbstractPluginIntegrationTest() {
...
+            (dependencies as ExtensionAware).extensions.create<Mine>("mine")
+        """)
+
+        withBuildScript("""
+            plugins {
+                `my-plugin`
+            }
+
+            dependencies {
+                println(mine.some + project.dependencies.mine.more)
```
bamboo: 👍

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/ScriptHandlerScopeTest.kt
```diff
@@ -42,7 +43,7 @@ class ScriptHandlerScopeTest {
     fun `can exclude modules from classpath dependency`() {
 
         val dependency = mock<ExternalModuleDependency>()
-        val dependencies = mock<DependencyHandler> {
+        val dependencies = mock<DependencyHandler>(arrayOf(ExtensionAware::class)) {
```
bamboo: ⭕️ Would it make sense to reuse `newDependencyHandlerMock` here?

Issue #1273, opened 2018-12-04
IDE resolver should cleanup its logging dir


eskatos:
implementing similar behavior as Gradle cleanup in the resolver embedded in the IDE.

eskatos:
Closing as fixed by #1281

PR #1281, opened 2018-12-05
Cleanup IDE resolver logs older than 7 days

eskatos:
See #1273

### DIFF
--- a/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/ResolverEventLogger.kt
+++ b/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/ResolverEventLogger.kt
@@ -76,7 +76,10 @@ object DefaultResolverEventLogger : ResolverEventLogger {
 
     private
     val outputFile by lazy {
-        File(outputDir(), "resolver-${timestampForFileName()}.log")
+        outputDir().let { logDir ->
+            cleanupLogDirectory(logDir)
+            logDir.resolve("resolver-${timestampForFileName()}.log")
+        }
     }
 
     private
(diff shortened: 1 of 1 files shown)


subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/ResolverEventLogger.kt
```diff
@@ -104,6 +107,21 @@ object DefaultResolverEventLogger : ResolverEventLogger {
     private
     fun now() = GregorianCalendar.getInstance().time
 
+    private
+    fun cleanupLogDirectory(logDir: File) {
+        val expiration = GregorianCalendar.getInstance().apply { add(Calendar.DAY_OF_MONTH, -cleanupAfterDays) }.time
+        logDir.listFiles { file ->
+            file.isFile && file.name.matches(resolverLogFilenameRegex) && Date(file.lastModified()).before(expiration)
+        }.forEach { it.delete() }
+    }
```
JLLeitschuh: In some rare cases, you'll have a system with a misconfigured time server that's date starts on 1 January 1970. If you subtract 7 days from 1 January 1970, do you end up with an error, or do you end up with all the log files deleted because of a time-based underflow?

If you're not concerned, feel free to ignore this.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/ResolverEventLogger.kt
```diff
@@ -76,7 +76,10 @@ object DefaultResolverEventLogger : ResolverEventLogger {
 
     private
     val outputFile by lazy {
-        File(outputDir(), "resolver-${timestampForFileName()}.log")
+        outputDir().let { logDir ->
+            cleanupLogDirectory(logDir)
```
bamboo: ❓ That would happen at least once per edited script type per project session, right?

I wonder if we shouldn't impose a harder limit to the amount of directory scanning here and say, cleanup once a week, for instance. What do you think?

eskatos: Once a day like [Gradle project caches cleanup](https://docs.gradle.org/current/userguide/directory_layout.html#dir:project_root:cache_cleanup) sounds good.

eskatos: I think this should do, PTAL

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/ResolverEventLogger.kt
eskatos: Right, the time based underflow would have the effect of cleaning up all log files right before creating the current one. I'm not too concerned, if you are in January the 1st 1970 you have other fish to fry than troubleshoot IJ editing kotlin build logic.

PR #1283, opened 2018-12-06
Let script dependencies resolver prioritise requests by recency

bamboo:


subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/concurrent/RessurectingThread.kt
```diff
@@ -0,0 +1,39 @@
...
+package org.gradle.kotlin.dsl.concurrent
+
+import kotlin.concurrent.thread
+
+
+/**
+ * A thread that comes back to life when [poke]d.
+ */
+internal
+class RessurectingThread(val name: String, val block: () -> Unit) {
```
JLLeitschuh: Clever way of solving this sort of problem!

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/concurrent/RessurectingThread.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    private
+    var thread: Thread? = null
+
+    /**
+     * Checks if the thread is alive, ressurecting it if it's not.
+     */
+    fun poke() = synchronized(this) {
+        if (thread?.isAlive != true) {
+            thread = thread(name = name, block = block)
```
JLLeitschuh: Do you want this to be a daemon thread so it dies when the main thread dies?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/KotlinBuildScriptModelRepository.kt
```diff
@@ -0,0 +1,96 @@
...
+        fetchKotlinBuildScriptModelFor(request)
+
+    private
+    val q = ConcurrentGroupingQueue<AsyncModelRequest> {
+        first.scriptFile == it.first.scriptFile && first.projectDir == it.first.projectDir
+    }
+
+    private
+    val processor = RessurectingThread("Kotlin DSL Resolver") {
+        while (true) {
```
JLLeitschuh: Do you want to do `while (!Thread.currentThread().isInterupted)` here?

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/resolver/ConcurrentGroupingQueueTest.kt
```diff
@@ -0,0 +1,76 @@
...
+        )
+    }
+
+    @Test
+    fun `#nextGroup honours timeout`() {
+
+        val subject = ConcurrentGroupingQueue<Int> { false }
+
+        thread {
+            Thread.sleep(150)
```
JLLeitschuh: Instead of using a hardcoded sleep, you might want to consider using something like this library with a waiter.

https://github.com/jhalterman/concurrentunit

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/ConcurrentGroupingQueue.kt
```diff
@@ -0,0 +1,101 @@
...
+        require(q.isNotEmpty())
+        val group = mutableListOf<T>()
+        q.iterator().run {
+            // the most recently pushed element
+            val mostRecent = next()
+            group.add(mostRecent)
+            remove()
+            // plus all elements superseded by it
+            for (next in this) {
+                if (mostRecent.supersedes(next)) {
```
JLLeitschuh: Maybe I'm completely misunderstanding, but wouldn't it be more performant to sort on insertion instead of filtering on every `take`?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/concurrent/RessurectingThread.kt
bamboo: Of course!

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/ConcurrentGroupingQueue.kt
bamboo: If I understand your question correctly, sorting already happens on insertion, thanks to `addFirst` the queue is kept sorted from most recent entry to least recent entry, that's why this method can be a simple scan on the tail of the queue to find superseded entries.

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/resolver/ConcurrentGroupingQueueTest.kt
bamboo: Thanks for the link, would you have a suggestion on how to improve this test to take advantage of that lib?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/KotlinBuildScriptModelRepository.kt
bamboo: Thanks for the suggestion, It might provide slightly better behaviour, yes. Notice that `nextGroup` already waits on a lock condition so the thread should already be nicely interruptible.

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/resolver/ConcurrentGroupingQueueTest.kt
JLLeitschuh: I'd reccomend doing something like this.

```kotlin
@Test
fun `#nextGroup honours push that happens in a different thread`() {

    val subject = ConcurrentGroupingQueue<Int> { false }
    val waiter = Waiter()

    thread {
        waiter.await()
        subject.push(42)
    }

    assertThat(
        subject.nextGroup(50),
        equalTo(emptyList())
    )
    waiter.resume()

    assertThat(
        subject.nextGroup(200),
        equalTo(listOf(42))
    )
}
```

JLLeitschuh: You may just be able to do this with a lock/semaphore though.

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/resolver/KotlinBuildScriptModelRepositoryTest.kt
```diff
@@ -0,0 +1,168 @@
...
+import kotlin.coroutines.Continuation
+
+
+class KotlinBuildScriptModelRepositoryTest {
+
+    @Test
+    fun `cancels older requests before returning response to most recent request`() = runBlockingWithTimeout {
+
+        // given:
+        val projectDir = File("/tmp/project")
```
eskatos: :x: this should use a `TemporaryFolder` rule or `FolderBasedTest`

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/resolver/ConcurrentGroupingQueueTest.kt
eskatos: +1 to finding a way to remove the hardcoded sleep

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/KotlinBuildScriptModelRepository.kt
```diff
@@ -0,0 +1,96 @@
...
+            process(group)
+        }
+    }
+
+    private
+    fun process(group: List<AsyncModelRequest>) {
+        val (mostRecentRequest, k) = group.first()
+        val result = runCatching { fetch(mostRecentRequest) }
+        resumeAll(supersededRequestsFrom(group), Result.success(null))
+        resume(k, result)
```
eskatos: ❓shouldn't this be resumed without `ignoringErrors {}` ?

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/resolver/KotlinBuildScriptModelRepositoryTest.kt
bamboo: I thought that would be adding unnecessary overhead for the files are never written to, they are just passed around as values. WDYT?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/KotlinBuildScriptModelRepository.kt
bamboo: No, we don't want an exception in the client to cause the resolver thread to quit.

eskatos: Right, I get it we don't bother about the cancelled requests. But I think that if this statement fails there's a real problem and we should log that, did I miss something?

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/resolver/KotlinBuildScriptModelRepositoryTest.kt
eskatos: Makes sense. What about capturing this more explicitly for the future reader then?

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/resolver/KotlinBuildScriptModelRepository.kt
bamboo: The logging should have happened on the client before propagating the exception. And if we need to diagnose some related error in the future, the `printStackTrace` should be enough (the output should appear in the IntelliJ log).

Notice this is the same behaviour as before.

eskatos: Got it now, thanks. SGTM

subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/resolver/ConcurrentGroupingQueueTest.kt
bamboo: I solved it with a `CountDownLatch` for simplicity.

Issue #1286, opened 2018-12-07
The `gradle-kotlin-dsl-extensions.jar` should be reproducible


eskatos:
### Expected Behavior
The `gradle-kotlin-dsl-extensions.jar` is reproducible to allow for downstream compilation tasks (when the `kotlin-dsl` plugin is applied) and script compilation outputs to be reused and relocatable.

### Current Behavior
Observed on CI where two linux agents produced a different jar file.

https://e.grdev.net/c/sprbsii6bnmjs/n5scix7scxwvy/task-inputs?toggledFileChanges=WyJ0eHFoZjNrY2FmdWFrLWZpbGUtaW5wdXQtMCIsImFzdzI1N2VtbXNtcXUtZmlsZS1pbnB1dC0wIl0#change-asw257emmsmqu-0-0

wolfs:
Those are not two linux agents, this is a Windows machine and my macbook: https://e.grdev.net/c/sprbsii6bnmjs/n5scix7scxwvy/infrastructure#operating-system

eskatos:
@wolfs thanks, updated the issue

wolfs:
Here is another instance of this problem from our CI agents, one on Windows and one one Linux: https://e.grdev.net/c/y66cprpcpn35u/2mohev2rc3yv2/task-inputs?toggled=WzJd&toggledFileChanges=WyI2ajZlNGlhdDV3Z3JlLWZpbGUtaW5wdXQtMCJd#change-6j6e4iat5wgre-0-0

wolfs:
I had a look at the jars, and it looks like line endings are the culprit. I guess we shouldn't use `appendln` here: https://github.com/gradle/kotlin-dsl/blob/9d4f2465571b3b8c30a3009951a85824018ca36f/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/codegen/ApiExtensionsGenerator.kt#L269-L293

wolfs:
I guess we should get rid of all usages of `appendln` in the generator source code.

PR #1288, opened 2018-12-08
Let generated extensions jar be reproducible

eskatos:
in order for script compilation classpath and `KotlinCompile` classpath in projects using the `kotlin-dsl` plugin (e.g. `buildSrc`) to be reproducible leading to more cache hits

See #1286


bamboo:
Required one more change done in 83e3686f2e7feb928beca9e98200f0203ba86662.

PR #1299, opened 2018-12-12
Let `PrintAccessors` emit the same code used for the extensions jar

bamboo:


subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/ProjectSchemaBasedTest.kt
```diff
@@ -0,0 +1,35 @@
...
+package org.gradle.kotlin.dsl.accessors
+
+import org.gradle.api.NamedDomainObjectContainer
+import org.gradle.api.reflect.TypeOf.parameterizedTypeOf
+
+import org.gradle.kotlin.dsl.fixtures.AbstractDslTest
+import org.gradle.kotlin.dsl.typeOf
+
+
+abstract class ProjectSchemaBasedTest : AbstractDslTest() {
```
eskatos: :o: I would name this type `ProjectSchemaBasedDslTest` so we know it's a `DslTest` at first sight

Issue #1297, opened 2018-12-12
Extract ktlint convention plugin into a dedicated repository


eskatos:

eskatos:
Also see follow up https://github.com/gradle/gradle/pull/8023

PR #1300, opened 2018-12-12
Move :plugins-experiments to their own repository

eskatos:
See #1297 

New home for the plugins https://github.com/gradle/kotlin-dsl-conventions

Published on the portal at https://plugins.gradle.org/plugin/org.gradle.kotlin-dsl.ktlint-convention


buildSrc/build.gradle.kts
```diff
@@ -3,7 +3,7 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 buildscript {
 
     val pluginsExperiments by extra(
```
bamboo: :x: Rename `pluginsExperiments` to something closer to the new dependency name.

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/accessors/AccessorsClassPath.kt
```diff
@@ -472,8 +472,8 @@ internal
 fun TypedProjectSchema.toCacheKeyString(): String =
     (cacheKeyPartsFor(extensions)
         + cacheKeyPartsFor(conventions)
-        //+ cacheKeyPartsFor(tasks) // TODO:accessors - add missing test case
-        //+ cacheKeyPartsFor(containerElements) // TODO:accessors - add missing test case
+        // + cacheKeyPartsFor(tasks) // TODO:accessors - add missing test case
```
bamboo: :x: this change doesn't seem relevant to this PR.

buildSrc/build.gradle.kts
eskatos: Fixed

subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/accessors/AccessorsClassPath.kt
eskatos: ktlint was upgraded and this is because of a new rule
it is a good thing I think

PR #1304, opened 2018-12-27
Let cache key derived from project schema include tasks and elements

bamboo:


subprojects/provider/src/test/kotlin/org/gradle/kotlin/dsl/accessors/ProjectSchemaTest.kt
```diff
@@ -199,21 +203,132 @@ class ProjectSchemaTest : TestWithClassPath() {
...
+                    entry<TaskContainer, DefaultTask>("clean")
+                )
+            ),
+            not(equalTo(cacheKeyFor(
+                tasks = listOf(
+                    entry<TaskContainer, Delete>("clean")
+                )
+            )))
+        )
+    }
```
JLLeitschuh: All of these assertions assert the lack of equality; any value in adding a test that covers checking that there is a case where equality is equal?
Is this already handled by `public type from jar is represented as Accessible`?

bamboo: We do have integration tests for caching which indirectly test the equality of cache keys but it makes sense to add a few unit tests as well. Done.

PR #1312, opened 2019-01-11
WIP: Type-safe accessors for precompiled script plugins

bamboo:
Not ready for review.


subprojects/plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/precompiled/ExtractPrecompiledScriptPluginPlugins.kt
```diff
@@ -0,0 +1,58 @@
...
+ */
+@CacheableTask
+open class ExtractPrecompiledScriptPluginPlugins : DefaultTask() {
+
+    @get:OutputDirectory
+    var outputDir = project.objects.directoryProperty()
+
+    @get:Internal
+    internal
+    lateinit var plugins: List<ScriptPlugin>
```
JLLeitschuh: Why not use a `ListProperty` here?

PR #1325, opened 2019-01-23
Kotlin 1.3.20

eskatos:
See https://github.com/JetBrains/kotlin/releases/tag/v1.3.20


subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/support/KotlinCompiler.kt
```diff
@@ -38,7 +38,7 @@ import org.jetbrains.kotlin.com.intellij.openapi.project.Project
 import org.jetbrains.kotlin.com.intellij.openapi.util.Disposer.dispose
 import org.jetbrains.kotlin.com.intellij.openapi.util.Disposer.newDisposable
 
-import org.jetbrains.kotlin.config.AnalysisFlag
+import org.jetbrains.kotlin.config.AnalysisFlags
```
JLLeitschuh: @eskatos Is this change at risk of making the Kotlin DSL incompatible with someone running an earlier version of the compiler?

PR #1339, opened 2019-01-31
Introduce resolver request correlation id

bamboo:
And make it possible to correlate events from the Kotlin DSL resolver log with the daemon log.

### DIFF
--- a/subprojects/tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinBuildScriptModelBuilder.kt
+++ b/subprojects/tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinBuildScriptModelBuilder.kt
@@ -96,7 +96,6 @@ object KotlinBuildScriptModelBuilder : ToolingModelBuilder {
         modelName == "org.gradle.kotlin.dsl.tooling.models.KotlinBuildScriptModel"
 
     override fun buildAll(modelName: String, modelRequestProject: Project): KotlinBuildScriptModel {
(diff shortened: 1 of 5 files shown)


subprojects/tooling-builders/src/main/kotlin/org/gradle/kotlin/dsl/tooling/builders/KotlinBuildScriptModelBuilder.kt
```diff
@@ -91,9 +95,22 @@ object KotlinBuildScriptModelBuilder : ToolingModelBuilder {
...
 
-    override fun buildAll(modelName: String, modelRequestProject: Project): KotlinBuildScriptModel =
-        scriptModelBuilderFor(modelRequestProject as ProjectInternal, requestParameterOf(modelRequestProject))
-            .buildModel()
+    override fun buildAll(modelName: String, modelRequestProject: Project): KotlinBuildScriptModel {
+        val logger = modelRequestProject.logger
+        val parameter = requestParameterOf(modelRequestProject)
+        try {
+            return kotlinBuildScriptModelFor(modelRequestProject, parameter).also {
+                println("$parameter => $it")
```
eskatos: ❓ Why use `println` and not the `logger` instance brought into scope a few lines above?

bamboo: `logger` output from the model builder doesn't end up in the daemon log (tried with `info` and `debug`). The variable is a leftover from a previous iteration. I'll remove it.

PR #1357, opened 2019-03-08
This may have a better Gradle kotlin version

i-walker:
I am struggling to run gradle tasks. With this, it worked out pretty good. 

Please review if there is another way without this file and actually adding it to the kotlin script
Check out this thread: https://stackoverflow.com/questions/27620262/sdk-location-not-found-define-location-with-sdk-dir-in-the-local-properties-fil

### Contributor Checklist
- [ ] Base the PR against the `develop` branch
- [ ] Provide integration tests to verify changes from a user perspective
- [ ] Provide unit tests to verify logic
- [ ] Ensure that tests pass locally: `./gradlew check --parallel`

### DIFF
--- a/samples/hello-android/local.properties
+++ b/samples/hello-android/local.properties
@@ -1,0 +1,2 @@
+// this file used to be auto-generated
+sdk.dir=/Users/.<Path to>./Android/sdk


i-walker:
I have a question before closing my PR. 
Why has the hello android sample a kts file in the root and a fully functional gradle kts in app, which should be the focus of the app right?

Are their any benifits to this

StefMa:
The Gradle Kotlin DSL team has only moved to the Kotlin DSL from an auto
generated "new project" from Android Studio. And if you create a new
project in AS,this is the default layout...

I am against these changes because the sample should work in AS. And this
will break it...

PR #1377, opened 2019-04-24
Update wrappers to Gradle 5.4

bamboo:


eskatos:
It also seems that all `gradlew` files lost their executable bit.

PR #1399, opened 2020-03-31
Update wrappers to Gradle 6.3

eskatos:
and add the wrapper validation github action
https://github.com/marketplace/actions/gradle-wrapper-validation


samples/hello-android/build.gradle.kts
```diff
@@ -7,7 +7,7 @@ buildscript {
     }
     dependencies {
         classpath("com.android.tools.build:gradle:3.1.3")
-        classpath(kotlin("gradle-plugin", version = "1.3.21"))
+        classpath(kotlin("gradle-plugin", version = "1.3.70"))
```
sschuberth: Would it make sense to go straight to version 1.3.71, @eskatos?

eskatos: This repository has checks that make sure the samples use the same kotlin version as the one embedded in used Gradle version in order to prevent forgetting updating the samples. 6.3 embeds 1.3.70.

Deviating from that would require more work to adjust the checks. Using 1.3.70 for now sounds ok to me.

